PR #18, opened 2018-12-18
Merge connection constructor with connect()

ColdenCullen:

### DIFF
--- a/aws_crt/mqtt.py
+++ b/aws_crt/mqtt.py
@@ -48,14 +48,7 @@ class Client(object):
 class Connection(object):
     __slots__ = ['_internal_connection', 'client', 'client_id', 'will']
 
-    def __init__(self, client, client_id):
-
-        assert isinstance(client, Client)
-
-        self.client = client
-        self.client_id = client_id
-
-    def connect(self,
+    def __init__(self, client, client_id,
             host_name, port,
             on_connect=_default_on_connect,
             on_disconnect=_default_on_disconnect,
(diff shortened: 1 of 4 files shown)


aws_crt/iot.py
```diff
@@ -85,13 +85,13 @@ def __init__(self, clientID, useWebsocket=False, cleanSession=True):
 
         self._username = None
         self._password = None
+        self._clientID = clientID
 
         self._elg = io.EventLoopGroup(1)
         self._bootstrap = io.ClientBootstrap(self._elg)
```
graebm: kinda unrelated, but PyLint on my machine is complaining about "io.EventLoopGroup" since neither "io" or "EventLoopGroup" is explicitly imported

graebm:
Also, since disconnect is a one-way door, we'd probably want to reject further publish and subscribe calls. The normal behavior is the queue them, but we should probably throw instead

aws_crt/iot.py
```diff
@@ -496,7 +496,8 @@ def _onDisconnectWrapper(return_code):
 
             self._client.tls_ctx = io.ClientTlsContext(self._tls_ctx_options)
```
graebm: oh, I meant to make that PyLint comment about this line here

aws_crt/mqtt.py
```diff
@@ -48,14 +48,7 @@ def __init__(self, bootstrap, tls_ctx = None):
 class Connection(object):
     __slots__ = ['_internal_connection', 'client', 'client_id', 'will']
```
graebm: I think `_internal_connection` is the only slot being used anymore

aws_crt/iot.py
graebm: I am a spaz. I keep meaning this comment to end up in mqtt.py

ColdenCullen:
I addressed the basic stuff, tomorrow we can talk about how we want to handle disconnect()

PR #23, opened 2018-12-21
Submodules and setup.

JonathanHenson:
I'll remove ninja dependency as a last nicety once we have everything else working.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,12 @@
+[submodule "aws-c-common"]
+	path = aws-c-common
+	url = git://github.com/awslabs/aws-c-common.git
+[submodule "aws-c-io"]
+	path = aws-c-io
+	url = git://github.com/awslabs/aws-c-io.git
+[submodule "aws-c-mqtt"]
+	path = aws-c-mqtt
+	url = git://github.com/awslabs/aws-c-mqtt.git
+[submodule "s2n"]
+	path = s2n
+	url = git://github.com/awslabs/s2n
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -1,8 +1,78 @@
...
 
+current_dir = os.path.dirname(os.path.realpath(__file__))
+
+build_dir = os.path.join(current_dir, 'deps_build')
+
+if not os.path.exists(build_dir):
+    os.mkdir(build_dir)
+
+os.chdir(build_dir)
+dep_install_path = os.getenv('AWS_C_INSTALL')
```
graebm: new users aren't going to have this environment variable. Should we just make a local install directory, like we make a local build directory?

.gitmodules
```diff
@@ -0,0 +1,12 @@
+[submodule "aws-c-common"]
+	path = aws-c-common
+	url = git://github.com/awslabs/aws-c-common.git
```
graebm: don't we want to checkout a tagged version?

JonathanHenson: it's set in the submodule itself

setup.py
```diff
@@ -1,8 +1,83 @@
...
+current_dir = os.path.dirname(os.path.realpath(__file__))
+
+build_dir = os.path.join(current_dir, 'deps_build')
+
+if not os.path.exists(build_dir):
+    os.mkdir(build_dir)
+
+os.chdir(build_dir)
+
+dep_install_path = ''
```
graebm: Probably we should set a full path, like AWS_C_INSTALL usually is?
And probably have it be a subdir of the build directory, instead of the build dir itself?
```
dep_install_path = os.path.join(os.getcwd(), 'install')
```

PR #25, opened 2018-12-27
Python2.7 fixes

JonathanHenson:


source/module.c
```diff
@@ -22,12 +22,22 @@
 
 #include <aws/mqtt/mqtt.h>
 
+#if PY_MAJOR_VERSION == 3
+#    define INIT_FN PyInit__aws_crt_python
+#    define UNICODE_GET_BYTES_FN PyUnicode_DATA
```
graebm: Do we want to use [PyUnicode_AsUTF8AndSize](https://docs.python.org/3/c-api/unicode.html#utf-8-codecs) so that we're sure we're getting utf-8?

source/module.c
```diff
@@ -22,12 +22,22 @@
...
+#    define UNICODE_GET_BYTE_LEN_FN PyUnicode_GET_LENGTH
```
graebm: This returns the number of "Code Points" in the string, not the number of bytes in the raw data

PR #27, opened 2018-12-31
Work fixing Windows CI builds.

JonathanHenson:

### DIFF
--- a/codebuild/common-windows.bat
+++ b/codebuild/common-windows.bat
@@ -2,6 +2,7 @@
 set CMAKE_ARGS=%*
 
 @"%SystemRoot%\System32\WindowsPowerShell\v1.0\powershell.exe" -NoProfile -InputFormat None -ExecutionPolicy Bypass -Command "iex ((New-Object System.Net.WebClient).DownloadString('https://chocolatey.org/install.ps1'))" && SET "PATH=%PATH%;%ALLUSERSPROFILE%\chocolatey\bin"
+echo "Installing python version: %PYTHON_PACKAGE% via choco"
 choco install %PYTHON_PACKAGE% -y
 call RefreshEnv.cmd
 


codebuild/common-macos.sh
```diff
@@ -6,6 +6,25 @@ set -e
...
+    pushd ./packages
+    # usually the existing package is too old for one of the others, so uninstall
+    # and reinstall from the cache
+    brew uninstall --ignore-dependencies $1
+    if [ ! -e $1*bottle*.tar.gz ]; then
+        brew install --build-bottle $1
+        brew bottle --json $1
+        brew uninstall --ignore-dependencies $1
+    fi
+    brew install $1*bottle*tar.gz
```
graebm: Should the `uninstall` be outside the if-block? Don't you want to uninstall/install regardless of whether the bottle exists in the cache?

codebuild/common-macos.sh
```diff
@@ -6,6 +6,25 @@ set -e
...
+    popd
+}
+
 install_from_brew openssl
 install_from_brew gdbm
 install_from_brew sqlite
```
graebm: where is the original `install_from_brew` function defined? I see your new function, but not what it was replacing. Was it something from Travis?

codebuild/common-windows.bat
```diff
@@ -1,14 +1,15 @@
 
-set CMAKE_ARGS=%*
 
 @"%SystemRoot%\System32\WindowsPowerShell\v1.0\powershell.exe" -NoProfile -InputFormat None -ExecutionPolicy Bypass -Command "iex ((New-Object System.Net.WebClient).DownloadString('https://chocolatey.org/install.ps1'))" && SET "PATH=%PATH%;%ALLUSERSPROFILE%\chocolatey\bin"
-choco install %PYTHON_PACKAGE% -y
+echo "Installing python version: %* via choco"
+choco install %* -y
+dir "C:\Program Files (x86)\Microsoft Visual Studio"
```
graebm: is this `dir` just for debugging?
do you still want it here?

codebuild/common-macos.sh
JonathanHenson: I had deleted it when we moved to letting python run the setup. I had to add it back to fix the build... this is just a rote copy from the original script justin wrote.

codebuild/common-macos.sh
JonathanHenson: this is the original, and it builds fine so, I vote for letting it be.

codebuild/common-windows.bat
JonathanHenson: removed now that I know why VS2017 wasn't working.

setup.py
```diff
@@ -26,20 +27,35 @@ def determine_cross_compile_string():
...
-        if vswhere_output != None:
-            for out in vswhere_output.split():
-                vs_version = out.decode('utf-8')
 
         if vs_version == None:
-            print('No version of MSVC compiler could be found!')
-            exit(1)
-
-        print('found MSVC compiler version: {}'.format(vs_version))
+            if os.path.exists('C:\\Program Files (x86)\\Microsoft Visual Studio\\2019'):
```
graebm: use `os.getenv("PROGRAMFILES(X86)")` instead of hardcoding "C:\\"

PR #31, opened 2019-01-26
Builds are now in release mode, purged unnecessary fluff from the sou…

JonathanHenson:
…rce distribution. Build output is now 800kb, and source distro is less than 4.5 MB.

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,3 +5,57 @@ Python bindings for the AWS Common Runtime
 ## License
 
 This library is licensed under the Apache 2.0 License. 
+
+## Building the Wheel
+
+This builds the wheel itself
+````bash
+<path to your python binary> setup.py sdist bdist_wheel
+````
+
+You'll want to do this for every version of Python:
+2.7m, 2.7mu, 3.4m, 3.5m, 3.6m, and 3.7m
+
+Don't worry, it will only do a full build on the first run. All the others will just reuse the artifacts.
+
+### Intel Linux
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -5,3 +5,57 @@ Python bindings for the AWS Common Runtime
...
+
+````bash
+<path to your python binary> -m twine upload --repository-url <the repo> dist/*
+````
+
+## Using From Your Python Application
+
+Normally, you just declare `aws_crt` as a dependency in your setup.py file.
+
+### Installing from pip
```
justinboswell: Shouldn't this come first, as it's the most user-friendly?

setup.py
```diff
@@ -212,7 +213,7 @@ def build_dependency(lib_name):
 
 setuptools.setup(
     name="aws_crt",
-    version="0.0.1",
+    version="0.2.2",
```
justinboswell: Shouldn't this be 0.2.0?

JonathanHenson: no, it should match the current tag, which will be cut as soon as I merge this.

README.md
JonathanHenson: er.... this isn't really for them, and publish then consume seems most logical for the layout. i just don't want the knowledge of how to publish this stuff to get lost in my head or on my bash history.

justinboswell: That's fair, though maybe it should go in PACKAGING.md? :)

Either way, I'm not super particular, it just jumped out at me.

setup.py
justinboswell: I should have looked, no worries.

PR #32, opened 2019-01-30
Release automation

JonathanHenson:


setup.py
```diff
@@ -213,7 +213,7 @@ def build_dependency(lib_name):
 
 setuptools.setup(
     name="aws_crt",
-    version="0.2.2",
+    version="0.2.1",
```
justinboswell: Downgrade?

codebuild/cd/test_version_exists.yml
```diff
@@ -0,0 +1,19 @@
+version: 0.2
+#this build spec assumes the ubuntu 14.04 trusty image
```
justinboswell: Might be worth a comment explaining how this is a canary used to gate the other builds

setup.py
JonathanHenson: tags now have to match version, I'll kick it up to 2.2 after this pr is merged

PR #33, opened 2019-01-31
Version 2.2 with extra safety check scripts.

JonathanHenson:

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -218,7 +218,7 @@ _aws_crt_python = setuptools.Extension(
 
 setuptools.setup(
     name="aws_crt",
-    version="0.2.2",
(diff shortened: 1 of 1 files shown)


continuous-delivery/sanity-check-test-pypi.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+CURRENT_TAG_VERSION=$(git describe --abbrev=0)
```
justinboswell: Do you want a set -e in here to ensure it dies when something fails?

JonathanHenson: good call.

PR #35, opened 2019-02-25
Add crypto, update module name and version.

JonathanHenson:
Sha256, Md5, SHA256 HMAC


awscrt/crypto.py
```diff
@@ -0,0 +1,47 @@
...
+    @staticmethod
+    def md5_new():
+        return Hash(native_handle=_aws_crt_python.aws_py_md5_new())
+
+    def update(self, to_hash):
+        _aws_crt_python.aws_py_hash_update(self._hash, to_hash)
+
+    def digest(self, truncate_to = 0):
+        tmp = _aws_crt_python.aws_py_hash_digest(self._hash, truncate_to)
+        return tmp
```
ColdenCullen: why tmp?

awscrt/crypto.py
```diff
@@ -0,0 +1,47 @@
...
+class Hash(object):
+    def __init__(self, native_handle):
+        self._hash = native_handle
+
+    @staticmethod
+    def sha256_new():
+        return Hash(native_handle=_aws_crt_python.aws_py_sha256_new())
+
+    @staticmethod
+    def md5_new():
```
ColdenCullen: I feel like these are confusing... We should probably remove the `_new`.

awscrt/crypto.py
```diff
@@ -0,0 +1,47 @@
...
+#  http://aws.amazon.com/apache2.0
+#
+# or in the "license" file accompanying this file. This file is distributed
+# on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+# express or implied. See the License for the specific language governing
+# permissions and limitations under the License.
+
+import _aws_crt_python
+
+class Hash(object):
```
ColdenCullen: The way this class is used is kind of confusing, can you document it please?

awscrt/crypto.py
JonathanHenson: doh! that was testing.

PR #45, opened 2019-04-18
Enabled TLS certs/key to be provided as bytes

justinboswell:

### DIFF
--- a/awscrt/io.py
+++ b/awscrt/io.py
@@ -37,13 +37,17 @@ class DefaultHostResolver(HostResolver):
         self._internal_host_resolver = _aws_crt_python.aws_py_io_host_resolver_new_default(max_hosts, elg._internal_elg)
 
 class ClientBootstrap(object):
-    __slots__ = ('elg', '_internal_bootstrap')
+    __slots__ = ('elg', 'host_resolver', '_internal_bootstrap')
 
-    def __init__(self, elg, host_resolver):
+    def __init__(self, elg, host_resolver=None):
         assert isinstance(elg, EventLoopGroup)
-        assert isinstance(host_resolver, HostResolver)
+        assert isinstance(host_resolver, HostResolver) or host_resolver is None
+
+        if host_resolver is None:
+            host_resolver = DefaultHostResolver(elg)
 
         self.elg = elg
+        self.host_resolver = host_resolver
         self._internal_bootstrap = _aws_crt_python.aws_py_io_client_bootstrap_new(self.elg._internal_elg, host_resolver._internal_host_resolver)
 
 class TlsVersion(IntEnum):
--- a/awscrt/io.py
+++ b/awscrt/io.py
@@ -41,12 +41,15 @@ class TlsVersion(IntEnum):
     DEFAULT = 128
 
 # force null termination at the end of buffer
+def byte_buf_null_terminate(buf):
+    if not buf.endswith(bytes([0])):
+        buf = buf + bytes([0])
+    return buf
+
 def byte_buf_from_file(filepath):
     with open(filepath, mode='rb') as fh:
         contents = fh.read()
-    if not contents.endswith(bytes([0])):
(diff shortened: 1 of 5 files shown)


awscrt/io.py
```diff
@@ -50,28 +64,49 @@ def __init__(self):
 
         self.min_tls_ver = TlsVersion.DEFAULT
 
-    def override_default_trust_store(self, ca_path, ca_file):
+    @classmethod
```
ColdenCullen: This doesn't belong here I don't think?

source/io.c
```diff
@@ -122,24 +122,30 @@ PyObject *aws_py_io_client_tls_ctx_new(PyObject *self, PyObject *args) {
     struct aws_allocator *allocator = aws_crt_python_get_allocator();
 
     int min_tls_version = AWS_IO_TLS_VER_SYS_DEFAULTS;
-    const char *ca_file = NULL;
     const char *ca_path = NULL;
+    const char *ca_buffer = NULL;
+    int ca_buffer_len = 0;
```
ColdenCullen: Py_ssize_t instead of int, otherwise you'll get some real gnar crashes later.

test.py
```diff
@@ -51,13 +51,15 @@ def on_receive_message(topic, message):
 tls_options = None
 if args.cert or args.key or args.root_ca:
     if args.cert:
-        assert(args.key)
-        tls_options = io.TlsContextOptions.create_client_with_mtls(args.cert, args.key)
+        assert (args.key)
```
ColdenCullen: remove parens, I believe. otherwise this evaluates to a tuple with one element in it, which is truthy regardless of the element's value

test.py
```diff
@@ -51,13 +51,15 @@ def on_receive_message(topic, message):
...
+        tls_options = io.TlsContextOptions.create_client_with_mtls_from_path(args.cert, args.key)
     else:
         tls_options = io.TlsContextOptions()
 
     if args.root_ca:
-        tls_options.override_default_trust_store(ca_path=None, ca_file=args.root_ca)
+        with open(args.root_ca, mode='rb') as ca:
+            rootca = ca.read()
+        tls_options.override_default_trust_store(rootca)
```
ColdenCullen: why not use `override_default_trust_store_from_path`?

source/io.c
justinboswell: Interesting, their docs literally say int:
```y# (read-only bytes-like object) [const char *, int]```

test.py
justinboswell: To exercise both paths and the null termination, and act as an example

test.py
justinboswell: I wondered about that, I didn't put them there, but I am definitely removing them.

awscrt/io.py
justinboswell: Why not?

ColdenCullen: Because it makes the method static, but the method is written like an instance method. If this is intended to be static, you should rename `self` to `cls`.

source/io.c
ColdenCullen: Note that we [define `PY_SSIZE_T_CLEAN`](https://github.com/awslabs/aws-crt-python/blob/master/source/module.h#L22) (as per [Parsing arguments and building values](https://docs.python.org/2.7/c-api/arg.html)):

> Starting with Python 2.5 the type of the length argument can be controlled by defining the macro PY_SSIZE_T_CLEAN before including Python.h. If the macro is defined, length is a Py_ssize_t rather than an int.

awscrt/io.py
justinboswell: @staticmethod does that, this is @classmethod

source/io.c
justinboswell: TIL

awscrt/io.py
justinboswell: Discussed offline, fixed.

PR #47, opened 2019-05-02
Http bindings

JonathanHenson:
Elasticurl.py, http bindings, and logging support.


awscrt/http.py
```diff
@@ -0,0 +1,153 @@
...
+from concurrent.futures import Future
+from awscrt.io import ClientBootstrap, TlsConnectionOptions, SocketOptions
+
+
+# Represents an Http connection to a remote endpoint. Everything in this class is non-blocking.
+class HttpClientConnection(object):
+    __slots__ = ('_bootstrap', '_tls_connection_options', '_on_connection_shutdown', '_native_handle')
+
+    # don't call me, I'm private
+    def __init__(self, bootstrap, on_connection_shutdown, tls_connection_options):
```
justinboswell: While there isn't full enforcement, you could do something like: https://stackoverflow.com/a/46459300 to be a bit more draconian

awscrt/http.py
```diff
@@ -0,0 +1,153 @@
...
+        if self._native_handle is not None:
+            return _aws_crt_python.aws_py_http_client_connection_is_open(self._native_handle)
+
+        return False
+
+    # Makes an Http request. When the headers from the response are received, the returned future will have a result.
+    # and request.response_headers will be filled in, and request.response_code will be available.
+    # After this future completes, you can get the result of request.response_completed,
+    # for the remainder of the response.
+    def make_request(self, request):
```
justinboswell: Note for future us: We should make sure this function is the same name across all languages. I think MakeRequest/make_request is probably the best, but we just need to check for this.

awscrt/http.py
```diff
@@ -0,0 +1,153 @@
...
+
+        for slot in self.__slots__:
+            setattr(self, slot, None)
+
+        self._bootstrap = bootstrap
+        self._tls_connection_options = tls_connection_options
+        self._on_connection_shutdown = on_connection_shutdown
+        self._native_handle = None
+
+    # Initiates a new connection to host_name and port using socket_options and tls_connection_options if supplied.
```
justinboswell: Any reason why we don't follow the PEP for docstrings: https://www.python.org/dev/peps/pep-0287/

Should allow for the best possible user experience in tools

source/http_client_connection.c
```diff
@@ -0,0 +1,551 @@
...
+    if (!error_code) {
+        py_connection->connection = connection;
+        capsule =
+            PyCapsule_New(py_connection, s_capsule_name_http_client_connection, s_http_client_connection_destructor);
+        py_connection->capsule = capsule;
+    } else {
+        aws_mem_release(py_connection->allocator, py_connection);
+    }
+
+    result = PyObject_CallFunction(py_connection->on_connection_setup, "(Ni)", capsule, error_code);
```
justinboswell: If there was an error, you just released the py_connection above, so this is going to fail. Maybe store on_connection_setup into a temporary before deletion?

source/http_client_connection.c
```diff
@@ -0,0 +1,551 @@
...
+    if (!py_connection->destructor_called) {
+        PyGILState_STATE state = PyGILState_Ensure();
+        PyObject *result = PyObject_CallFunction(py_connection->on_connection_shutdown, "(i)", error_code);
+        Py_XDECREF(result);
+        PyGILState_Release(state);
+    } else {
+        aws_mem_release(py_connection->allocator, py_connection);
+    }
+
+    Py_DECREF(py_connection->on_connection_shutdown);
```
justinboswell: Same here, you just deleted py_connection, you need to invoke on_connection_shutdown before deletion or cache it.

source/http_client_connection.c
```diff
@@ -0,0 +1,551 @@
...
+            &host_name_len,
+            &port_number,
+            &py_socket_options,
+            &tls_conn_options_capsule)) {
+        PyErr_SetNone(PyExc_ValueError);
+        goto error;
+    }
+
+    if (!bootstrap_capsule || !PyCapsule_CheckExact(bootstrap_capsule)) {
+        PyErr_SetNone(PyExc_ValueError);
```
justinboswell: I think you want ```PyErr_SetString(PyExc_ValueError, "bootstrap is invalid");``` for each of these arguments, otherwise this is very hard to diagnose for the user.

source/http_client_connection.c
```diff
@@ -0,0 +1,551 @@
...
+    (void)internal_stream;
+
+    struct py_http_stream *stream = user_data;
+
+    PyGILState_STATE state = PyGILState_Ensure();
+
+    Py_ssize_t data_len = (Py_ssize_t)buf->capacity - buf->len;
+
+    PyObject *mv = PyMemoryView_FromMemory((char *)(buf->buffer + buf->len), data_len, PyBUF_WRITE);
+    PyObject *result = PyObject_CallFunction(stream->on_read_body, "(O)", mv);
```
justinboswell: So this version is going to require the callback to be called n+1 times, which is unlike how we handle it elsewhere. Can we not do it the same way for python?

source/http_client_connection.c
```diff
@@ -0,0 +1,551 @@
...
+static void s_on_incoming_header_block_done(struct aws_http_stream *internal_stream, bool has_body, void *user_data) {
+    (void)has_body;
+
+    struct py_http_stream *stream = user_data;
+
+    PyGILState_STATE state = PyGILState_Ensure();
+    int response_code = 0;
+    aws_http_stream_get_incoming_response_status(internal_stream, &response_code);
+    PyObject *result =
+        PyObject_CallFunction(stream->on_incoming_headers_received, "(Oi)", stream->received_headers, response_code);
```
justinboswell: Why not pass along has_body?

awscrt/http.py
JonathanHenson: I asked the python team which they hated least and this was their advisement.

source/http_client_connection.c
JonathanHenson: not unless we have python return a tuple

JonathanHenson: I'll play with it in a bit and see what I can do

source/http_client_connection.c
```diff
@@ -0,0 +1,560 @@
...
+    }
+
+    if (!bootstrap_capsule || !PyCapsule_CheckExact(bootstrap_capsule)) {
+        PyErr_SetString(PyExc_ValueError, "bootstrap is invalid");
+        goto error;
+    }
+
+    if (tls_conn_options_capsule && tls_conn_options_capsule != Py_None &&
+        !PyCapsule_CheckExact(tls_conn_options_capsule)) {
+        PyErr_SetNone(PyExc_ValueError);
```
justinboswell: Same as bootstrap, this needs to be more descriptive

source/http_client_connection.c
```diff
@@ -0,0 +1,560 @@
...
+    }
+
+    if (!(host_name && py_socket_options && on_connection_setup && on_connection_shutdown)) {
+        PyErr_SetString(PyExc_ValueError, "invalid arguments");
+        goto error;
+    }
+
+    struct aws_client_bootstrap *bootstrap = PyCapsule_GetPointer(bootstrap_capsule, s_capsule_name_client_bootstrap);
+    if (!bootstrap) {
+        PyErr_SetNone(PyExc_ValueError);
```
justinboswell: This should be more descriptive

source/http_client_connection.c
```diff
@@ -435,14 +448,14 @@ PyObject *aws_py_http_client_connection_make_request(PyObject *self, PyObject *a
...
-        PyErr_SetNone(PyExc_ValueError);
+        PyErr_SetString(PyExc_ValueError, "http connection capsule is invalid");
         goto clean_up_stream;
     }
 
     py_connection = PyCapsule_GetPointer(http_connection_capsule, s_capsule_name_http_client_connection);
 
     if (!py_http_request || !on_stream_completed || !on_incoming_headers_received) {
-        PyErr_SetNone(PyExc_ValueError);
+        PyErr_SetString(PyExc_ValueError, "invalid arguments");
```
justinboswell: I kinda feel like we should split this up so the error message is more clear

source/http_client_connection.c
```diff
@@ -137,7 +140,7 @@ PyObject *aws_py_http_client_connection_create(PyObject *self, PyObject *args) {
     }
 
     if (!(host_name && py_socket_options && on_connection_setup && on_connection_shutdown)) {
-        PyErr_SetNone(PyExc_ValueError);
+        PyErr_SetString(PyExc_ValueError, "invalid arguments");
```
justinboswell: Could we split this up to make the error more clear?

awscrt/http.py
```diff
@@ -0,0 +1,174 @@
...
+        for slot in self.__slots__:
+            setattr(self, slot, None)
+
+        self._bootstrap = bootstrap
+        self._tls_connection_options = tls_connection_options
+        self._on_connection_shutdown = on_connection_shutdown
+        self._native_handle = None
+
+    @staticmethod
+    def new_connection(bootstrap, host_name, port, socket_options, on_connection_shutdown, tls_connection_options):
```
graebm: `tls_connection_options=None`

awscrt/http.py
```diff
@@ -0,0 +1,174 @@
...
+        on_connection_shutdown takes a single argument of int type, to specify the shutdown reason.
+
+        returns a future where the result is a new instance to HttpClientConnection, once the connection has completed
+        and is ready for use.
+        """
+        assert tls_connection_options is None or isinstance(tls_connection_options, TlsConnectionOptions)
+        assert host_name is not None
+        assert port is not None
+        assert socket_options is not None and isinstance(socket_options, SocketOptions)
+        assert on_connection_shutdown is not None
```
graebm: Seems like we should allow `on_connection_shutdown` to be None
Which means it should also get a default `on_connection_shutdown=None` in the fn declaration

awscrt/http.py
```diff
@@ -0,0 +1,174 @@
...
+
+        except Exception as e:
+            future.set_exception(e)
+
+        return future
+
+    def close(self):
+        """
+        Closes the connection, if you hold a reference to this instance of HttpClientConnection, on_connection_shutdown
+        will be invoked upon completion of the connection close.
```
graebm: the "if you hold a reference to this instance of HttpClientConnection" part of this comment seems unnecessary. If you're calling a member function, you have a reference

awscrt/http.py
```diff
@@ -0,0 +1,174 @@
...
+
+
+class OutgoingHttpBodyState(IntEnum):
+    InProgress = 0
+    Done = 1
+
+
+class HttpRequest(object):
+    """
+    Represents an HttpRequest to pass to HttpClientConnection.make_request(). path_and_query is the path and query portion
```
graebm: I don't think we need a "builder/options" class like this in Python. Python handles default-argument and named-arguments really well, so it's not a bad idea to just pile args into the make_request(...) function

PR #51, opened 2019-05-09
Fixes for logger getting cleaned up too early, also handles null term…

JonathanHenson:
…inated buffers properly.


source/io.c
```diff
@@ -259,6 +202,16 @@ PyObject *aws_py_io_client_tls_ctx_new(PyObject *self, PyObject *args) {
     if (certificate_buffer && private_key_buffer && certificate_buffer_len > 0 && private_key_buffer_len > 0) {
         struct aws_byte_cursor cert = aws_byte_cursor_from_array(certificate_buffer, certificate_buffer_len);
         struct aws_byte_cursor key = aws_byte_cursor_from_array(private_key_buffer, private_key_buffer_len);
+        
+       /* adding the null terminator in python land screws up the byte buf apis. We never consider the null terminator
```
justinboswell: Ah, fuck, that's my fault. How did this ever work ever?

source/module.c
```diff
@@ -29,6 +29,53 @@
...
+        .level = log_level,
+        .file = NULL,
+        .filename = NULL,
+    };
+
+    Py_ssize_t stdout_len = (Py_ssize_t)strlen("stdout");
+
+    Py_ssize_t cmp_len = file_path_len > stdout_len ? stdout_len : file_path_len;
+
+    if (!memcmp("stdout", file_path, (size_t)cmp_len)) {
```
justinboswell: Should we allow these to be case-insensitive?

JonathanHenson: I don't think the extra branches are worth it, and it's written in lowercase in python anyways.

source/io.c
JonathanHenson: it worked fine till we started loading certs on apple, and it did not like that NULL at all

PR #57, opened 2019-08-08
Server create&destroy

TingDaoK:
*Issue #, if available:*

*Description of changes:*


awscrt/http.py
```diff
@@ -144,6 +152,73 @@ def on_incoming_headers_received(headers, response_code, has_body):
...
+            internal_conn_options_handle = None
+        server = HttpServer(bootstrap, on_incoming_connection,
+                            on_destroy_complete, tls_connection_options)
+        server._native_handle = _aws_crt_python.aws_py_http_server_create(
+            bootstrap._internal_bootstrap, on_incoming_connection, on_destroy_complete, host_name, port, socket_options, internal_conn_options_handle)
+        return server
+
+    def release(self):
+        """
+        release the server, no more connections will be accepted, when the server finishes destroy process, the on_destroy_complete will be invoked
```
graebm: maybe rename this `close()`
In C, release() is a combination "shut this thing down and release my refcount" call.
In python, users don't think about refcounts so lets use simple terms

Also, the docstring should mention the fact that a future is being returned. Also, we might want to rename remove the `on_destroy_complete` callback that's passed into the constructor, since it doesn't always fire, and we provide the user this future if they want to learn about "shutdown complete"

source/http_server.c
```diff
@@ -0,0 +1,276 @@
...
+
+static void s_on_destroy_complete(void *user_data) {
+    struct py_http_server *py_server = user_data;
+    PyGILState_STATE state = PyGILState_Ensure();
+    py_server->destroy_complete = true;
+    PyObject *on_destroy_complete_cb = py_server->on_destroy_complete;
+    PyObject *result = NULL;
+    if (!py_server->destructor_called && on_destroy_complete_cb) {
+        result = PyObject_CallFunction(on_destroy_complete_cb, "(N)", py_server->capsule);
+        Py_DECREF(on_destroy_complete_cb);
```
graebm:
```
    if (result) {
        Py_DECREF(result);
    } else {
        PyErr_WriteUnraisable(PyErr_Occurred());
    }
```

Some things about [PyObject_CallFunction()](https://docs.python.org/3/c-api/object.html#c.PyObject_CallFunction):
1) If `result` is non-null, you need to decref it
2) If `result` IS null then an exception has occurred. Either the callback couldn't be called, or an exception was raised within the callback that didn't get caught. There's not much we can do now that we're back in C-Town, but we can use [PyErr_WriteUnraisable()](https://docs.python.org/3/c-api/exceptions.html#c.PyErr_WriteUnraisable) to print a msg to stderr where the user might see it

source/http_server.c
```diff
@@ -0,0 +1,276 @@
...
+    }
+
+    if (!bootstrap_capsule || !PyCapsule_CheckExact(bootstrap_capsule)) {
+        PyErr_SetString(PyExc_ValueError, "bootstrap is invalid");
+        goto error;
+    }
+
+    if (tls_conn_options_capsule && tls_conn_options_capsule != Py_None &&
+        !PyCapsule_CheckExact(tls_conn_options_capsule)) {
+        PyErr_SetString(PyExc_ValueError, "tls connection options is invalid");
```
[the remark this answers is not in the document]
TingDaoK: I just deleted it, the check is made by the PyCapsule_GetPointer function right?

awscrt/http.py
```diff
@@ -144,6 +145,84 @@ def on_incoming_headers_received(headers, response_code, has_body):
...
+class HttpServer(object):
+    """
+    Represents an Http server. Everything in this class is non-blocking.
+    """
+    __slots__ = ('_bootstrap', '_tls_connection_options', '_on_incoming_connection',
+                 '_on_destroy_complete', '_native_handle', 'destroy_complete')
+
+    # don't call me, I'm private
+    def __init__(self, bootstrap, on_incoming_connection, on_destroy_complete, tls_connection_options):
+        assert isinstance(bootstrap, ServerBootstrap)
```
graebm: Why not use the default constructor? I don't see the advantage of forcing users to call the static `HttpServer.new_server(...)` vs normal `HttpServer(...)` python construction

awscrt/http.py
```diff
@@ -144,6 +145,84 @@ def on_incoming_headers_received(headers, response_code, has_body):
...
+        server._native_handle = _aws_crt_python.aws_py_http_server_create(
+            bootstrap._internal_bootstrap, on_incoming_connection, on_destroy_complete, host_name, port, socket_options, internal_conn_options_handle)
+        
+        return server
+
+    def close(self):
+        """
+        close the server, no more connections will be accepted, a future object will be returned, and if the close process finishes
+        the future result or exception will be set.
+        """
```
graebm: "... **when** the close process finishes..."
Futures MUST be set to something eventually. It is a crime to never set it to anything.

awscrt/http.py
```diff
@@ -144,6 +145,84 @@ def on_incoming_headers_received(headers, response_code, has_body):
...
+        try:
+            if self._native_handle is not None:
+                _aws_crt_python.aws_py_http_server_release(self._native_handle)
```
graebm: Is it possible for `self._native_handle` to be None?

Also, we should make `close()` idempotent (safe for user to call multiple times).

awscrt/http.py
```diff
@@ -144,6 +145,84 @@ def on_incoming_headers_received(headers, response_code, has_body):
...
+
+        for slot in self.__slots__:
+            setattr(self, slot, None)
+
+        self._bootstrap = bootstrap
+        self._tls_connection_options = tls_connection_options
+        self._on_incoming_connection = on_incoming_connection
+        self._on_destroy_complete = on_destroy_complete
+        self._native_handle = None
+        self.destroy_complete = Future()
```
graebm: private variables should start with _

awscrt/http.py
```diff
@@ -144,6 +145,82 @@ def on_incoming_headers_received(headers, response_code, has_body):
...
+        @param socket_options: awscrt.io.SocketOptions for the server's listening socket. Required
+        @param on_incoming_connection: Callback with signature (connection: HttpConnection.native_handle, error_code: int) Required
+        @param bootstrap: awscrt.io.ServerBootstrap. Required
+        @param tls_connection_options: awscrt.io.TlsConnectionOptions, for TLS connection
+        """
+        assert tls_connection_options is None or isinstance(tls_connection_options, TlsConnectionOptions)
+        assert host_name is not None
+        assert port is not None
+        assert socket_options is not None and isinstance(
+            socket_options, SocketOptions)
```
graebm: don't need to check `not None`, because None can't be an instance of SocketOptions

awscrt/http.py
```diff
@@ -144,6 +145,82 @@ def on_incoming_headers_received(headers, response_code, has_body):
...
+        assert on_incoming_connection is not None
+        if tls_connection_options is not None:
+            internal_conn_options_handle = tls_connection_options._internal_tls_conn_options
+        else:
+            internal_conn_options_handle = None
+
+        def on_destroy_complete(server_native_handle):
+            server.destroy_complete.set_result("server destroy finished!")
+
```
graebm: `set_result(None)` is fine when there's no value needed

source/http_server.c
```diff
@@ -0,0 +1,228 @@
...
+    struct aws_server_bootstrap *bootstrap = PyCapsule_GetPointer(bootstrap_capsule, s_capsule_name_server_bootstrap);
+    if (!bootstrap) {
+        PyErr_SetString(PyExc_ValueError, "the bootstrap capsule has an invalid pointer");
+        goto error;
+    }
+
+    struct aws_tls_connection_options *connection_options = NULL;
+
+    if (tls_conn_options_capsule != Py_None) {
+        connection_options = PyCapsule_GetPointer(tls_conn_options_capsule, s_capsule_name_tls_conn_options);
```
graebm: Not checking whether GetPointer succeeds (it will fail if they passed in the wrong type)

source/http_server.c
```diff
@@ -0,0 +1,228 @@
...
+    }
+
+    if (!PyCallable_Check(on_destroy_complete)) {
+        PyErr_SetString(PyExc_TypeError, "on_destroy_complete is invalid");
+        goto error;
+    }
+
+    struct aws_server_bootstrap *bootstrap = PyCapsule_GetPointer(bootstrap_capsule, s_capsule_name_server_bootstrap);
+    if (!bootstrap) {
+        PyErr_SetString(PyExc_ValueError, "the bootstrap capsule has an invalid pointer");
```
graebm: PyCapsule already sets an exception, no need to call PyErr_SetString

source/http_client_connection.c
```diff
@@ -176,40 +176,8 @@ PyObject *aws_py_http_client_connection_create(PyObject *self, PyObject *args) {
...
-
-    PyObject *keep_alive_timeout = PyObject_GetAttrString(py_socket_options, "keep_alive_timeout_secs");
-    if (keep_alive_timeout) {
-        socket_options.keep_alive_timeout_sec = (uint16_t)PyLong_AsLong(keep_alive_timeout);
-    }
-
-    PyObject *keep_alive_max_probes = PyObject_GetAttrString(py_socket_options, "keep_alive_max_probes");
-    if (keep_alive_timeout) {
-        socket_options.keep_alive_max_failed_probes = (uint16_t)PyLong_AsLong(keep_alive_max_probes);
+    if(!aws_socket_options_init_from_py(&socket_options, py_socket_options)){
```
graebm: move the AWS_ZERO_STRUCT(*socket_options) into the init() function

source/http_server.c
```diff
@@ -0,0 +1,228 @@
...
+    struct aws_socket_endpoint endpoint;
+    AWS_ZERO_STRUCT(endpoint);
+
+    snprintf(endpoint.address, host_name_len, "%s", host_name);
+    endpoint.port = port_number;
+    options.socket_options = &socket_options;
+    options.endpoint = &endpoint;
+    options.on_incoming_connection = s_on_incoming_connection;
+    options.on_destroy_complete = s_on_destroy_complete;
+    py_server->server = aws_http_server_new(&options);
```
graebm: not checking results of aws_http_server_new()

source/http_server.c
```diff
@@ -0,0 +1,228 @@
...
+    PyObject *capsule = NULL;
+    if (py_server->server) {
+        /* success */
+        capsule = PyCapsule_New(py_server, s_capsule_name_http_server, s_http_server_destructor);
```
graebm: not checking results of PyCapsule_New()

source/http_server.c
```diff
@@ -0,0 +1,228 @@
...
+    struct aws_socket_options socket_options;
+    AWS_ZERO_STRUCT(socket_options);
+    if (!aws_socket_options_init_from_py(&socket_options, py_socket_options)) {
+        goto error;
+    }
+
+    Py_INCREF(on_incoming_connection);
+    py_server->on_incoming_connection = on_incoming_connection;
+
+    Py_INCREF(on_destroy_complete);
```
graebm: you might want to wait until you're super sure that everthing has succeeded before doing Py_INCREF.

if you do it too early in the function, and errors might occur later, you have to worry whether or not to DECREF the object

source/http_server.c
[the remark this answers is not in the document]
graebm: The check is made by PyCapsule_GetPointer(), but you're not checking whether or not it returned NULL

awscrt/http.py
TingDaoK: It is totally safe to call it multiple times.

source/http_server.c
TingDaoK: I guess if(py_server->server) check the result of that new function

source/http_server.c
```diff
@@ -0,0 +1,228 @@
...
+        PyObject *result = PyObject_CallFunction(on_destroy_complete_cb, "(N)", py_server->capsule);
+        if(result){
+            Py_DECREF(result);
+        }
+        else{
+            PyErr_WriteUnraisable(PyErr_Occurred());
+        }
+        Py_DECREF(py_server->on_destroy_complete);
+        
+        Py_DECREF(py_server->capsule);
```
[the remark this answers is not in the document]
TingDaoK: if the destructor was called, the ref_count of py_server->capsule should be already 0.
But py_server->on_destroy_complete do need to be DECREF

awscrt/http.py
```diff
@@ -144,31 +143,94 @@ def on_incoming_headers_received(headers, response_code, has_body):
...
+    def close(self):
+        """
+        close the server, no more connections will be accepted, a future object will be returned, and when the close process finishes
+        the future result or exception will be set.
+        """
+        try:
+            _aws_crt_python.aws_py_http_server_release(self._native_handle)
+        except Exception as e:
+            self._destroy_complete.set_exception(e)
```
graebm: I don't think we should use try/except here. We should probably just let the exception go uncaught and let the user handle it. The only cause for exception is this function somehow running on an object that isn't actually a Server capsule, right?

source/http_server.c
```diff
@@ -0,0 +1,230 @@
...
+PyObject *aws_py_http_server_release(PyObject *self, PyObject *args) {
+    (void)self;
+
+    PyObject *server_capsule = NULL;
+
+    if (PyArg_ParseTuple(args, "O", &server_capsule)) {
+        if (server_capsule != Py_None) {
+            struct py_http_server *py_server = PyCapsule_GetPointer(server_capsule, s_capsule_name_http_server);
+            if(!py_server){
+                Py_RETURN_NONE;
```
graebm: In this case, shouldn't we `return NULL;` and cause an exception to be raised in python-land?

TingDaoK: I made this change to request handler branch, but forget to update here...

PR #65, opened 2019-09-05
H1B (2nd try)

graebm:
- Update dependencies to latest
- Revamp HTTP API
  - Base classes for Connection, Stream, Request, in preparation for server stuff
  - `HttpRequest` is its own class, distinct from `HttpClientStream`.
  - `HttpHeaders` is a convenience class for accessing headers pythonically
- `aws_input_stream` implementation for streaming from Python's I/O classes
- `aws_raise_py_error()` is new, translates Python exceptions into `aws_raise_error(AWS_ERROR_XYZ)` 
- Unit tests!
- trivial, but have just one http.h file, with many http_xyz.c files. 

I still need to add more unit tests and documentation (some was deleted), but wanted to get this bug chunk of work out there for review 


source/http_message.c
```diff
@@ -0,0 +1,235 @@
...
+#include <aws/io/stream.h>
+
+static const char *s_capsule_name_http_message = "aws_http_message";
+
+/**
+ * Bind aws_http_message to HttpRequest/HttpResponse.
+ *
+ * This binding acts differently than most. Instead of keeping the native type in
+ * sync with the Python type at all times, we lazily sync their contents only when we must.
+ * We do this because it seems cruel to make Python developers them access headers via the C API,
```
justinboswell: s/seems/is/ :)

source/http_message.c
```diff
@@ -0,0 +1,235 @@
...
+            goto error;
+        }
+
+        aws_http_message_set_body_stream(request->native, request->native_body_stream);
+    }
+
+    return capsule;
+
+error:
+    if (capsule) {
```
justinboswell: This needs (because the proxy can fail):
```cpp
if(request->native) { 
    aws_http_message_destroy(request->native);
}
```

source/http_stream.c
```diff
@@ -51,148 +50,163 @@ struct aws_http_stream *aws_py_get_http_stream(PyObject *stream) {
...
-    PyObject *result = PyObject_CallFunction(stream->on_read_body, "(O)", mv);
-    /* the return from python land is a tuple where the first argument is the body state
-     * and the second is the amount written */
-    PyObject *state_code = PyTuple_GetItem(result, 0);
-    enum aws_http_outgoing_body_state body_state = (enum aws_http_outgoing_body_state)PyIntEnum_AsLong(state_code);
-    PyObject *written_obj = PyTuple_GetItem(result, 1);
-    long written = (long)PyLong_AsLong(written_obj);
-    Py_XDECREF(result);
-    Py_XDECREF(mv);
+        bool append_success = PyList_Append(stream->received_headers, name_value_pair) == 0;
```
justinboswell: Interesting. Just out of curiosity, why not buffer in native and deliver all of the headers at once when the block is done, and not go over the VM barrier for each chunk?

source/http_stream.c
```diff
@@ -51,148 +50,163 @@ struct aws_http_stream *aws_py_get_http_stream(PyObject *stream) {
...
-    (void)internal_stream;
+static void s_on_stream_complete(struct aws_http_stream *native_stream, int error_code, void *user_data) {
+    (void)native_stream;
     struct http_stream_binding *stream = user_data;
 
+    /*************** GIL ACQUIRE ***************/
     PyGILState_STATE state = PyGILState_Ensure();
 
-    PyObject *result = PyObject_CallFunction(stream->on_stream_completed, "(i)", error_code);
+    /* Set this just in case callback fires before aws_http_connection_make_request() has even returned */
```
justinboswell: If that's a possibility, we have unleashed Zalgo and we should not do that.

source/http_stream.c
```diff
@@ -201,137 +215,84 @@ PyObject *aws_py_http_client_stream_new(PyObject *self, PyObject *args) {
...
 
-    struct aws_http_stream *http_stream = aws_http_stream_new_client_request(&request_options);
-    aws_array_list_clean_up(&headers);
+    Py_DECREF(capsule);
+    Py_RETURN_NONE;
 
-    if (!http_stream) {
-        goto clean_up_headers;
+error:
+    if (capsule) {
```
justinboswell: Don't we need to decref on the proxy here too, if failure happens after we incref?

source/module.h
```diff
@@ -51,7 +51,18 @@ void PyErr_SetAwsLastError(void);
...
-PyObject *aws_py_memory_view_from_byte_buffer(struct aws_byte_buf *buf, int flags);
+/**
+ * Raise an AWS error corresponding to the current Python error.
+ *
+ * Prints the current Python error to stderr and clears the Python error indicator.
+ * Finds an AWS error code corresponding to the current Python error (fallback is AWS_ERROR_UNKNOWN).
+ * Invokes aws_raise_error() with that error code and always returns AWS_OP_ERR.
+ *
+ * The Python error indicator MUST be set and the GIL MUST be held when calling this function. */
+int aws_raise_py_error(void);
```
justinboswell: ```aws_py_raise_error```

awscrt/http.py
```diff
@@ -100,92 +109,147 @@ def on_connection_setup(binding, error_code):
...
-        self.response_headers_received = Future()
-        self._stream = None
-        self.response_headers = None
-        self.response_code = None
-        self.has_response_body = False
+        self.path = path
+        self._binding = _awscrt.http_request_new(self, self._body_stream)
+
+
+class HttpHeaders(object):
```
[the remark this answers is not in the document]
graebm: The idea was that users could always access the `HttpHeaders.map` for a normal Pythonic experience, and the functions on this class were just helpers for common operations. This was inspired by [golang's HttpHeader type](https://golang.org/pkg/net/http/#Header) which is just a map with extra functions attached it.

I could learn more about making something natively iterable in python. That would open the door to binding this more tightly to C datastrcture, instead of lazily updating the C datastructure...

elasticurl.py
```diff
@@ -121,113 +115,80 @@ def print_header_list(headers):
...
 
 # make the request
-request = connection.make_request(method, uri_str, outgoing_headers, on_outgoing_body, on_incoming_body)
-request.response_headers_received.add_done_callback(response_received_cb)
+stream = connection.request(request, response_received_cb, on_incoming_body)
 
 # wait until the full response is finished
-response_finished = request.response_completed.result(timeout=10)
-request = None
+stream.complete_future.result(timeout=10)
```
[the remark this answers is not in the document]
graebm: I'll remove the timeout for now

source/http_message.c
graebm: nope! I have stumbled across a glorious python-only error handling technique:
Once you've created a capsule, it's destructor is guaranteed to run.
So I create a capsule around the binding struct as soon as possible. Then everything else that might need a teardown goes into the binding struct. If something goes wrong during initialization the capsule gets GC'd, the destructor runs, and everything gets cleaned up.

source/http_stream.c
graebm: That is exactly what we're doing

graebm: You make an excellent point. You could argue that's how the aws-c-http API should work, buffering response headers for you until they're all ready. I'll think about this...

source/http_stream.c
graebm: This is that thing we chatted about where it seems like we might want some kind of `.activate()` function on the stream that users need to call it kick it off. I tried really hard to do things RAII in aws-c-http, not having objects in a "half-initialized" state, but we end up in situations like this...

If we're going to address this in aws-c-http, we need to set aside time to do so.

source/http_stream.c
graebm: same thing where capsule destructor will do the cleanup

source/module.h
graebm: I do not like this naming convention where we start everything with `aws_py_` just because it's in a crt header file. This helper function is only for use in C, and it's a variant on the existing function `aws_raise_error()`, so I named it like the other function.

We were already doing this with functions like `PyErr_SetAwsLastError()` which is a variant on the existing `PyErr_SetString()`

source/http_message.c
justinboswell: Maybe a comment then? That was confusing to read.

source/module.h
graebm: renaming this one since they're so similar

source/http_stream.c
graebm: Adding a task to address this. Neither of us like the `stream.activate()` solution, but neither of us can think of anything else bulletproof

source/http_stream.c
graebm: did it
we also discussed making an `aws_http_headers` datastructure that we could use in cases like this, outside of `aws_http_message`

awscrt/http.py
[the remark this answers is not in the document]
graebm: iterable
and so much more...

awscrt/http.py
```diff
@@ -74,7 +75,7 @@ def new(cls, host_name, port, socket_options=SocketOptions(), tls_connection_opt
         future = Future()
         try:
             assert isinstance(bootstrap, ClientBootstrap) or bootstrap is None
-            assert host_name
+            assert isinstance(host_name, str) and host_name
```
justinboswell: Doesn't ```isinstance()``` demand that ```host_name``` is not ```None```?

graebm: `isinstance()` will return `False` for any `None`
the `and host_name` check if to ensure it's not ""

graebm: I kind of feel like we should have a `is_nonempty_str(x)` helper, but I'd need to check a ton of existing code to see who should be using that

justinboswell: I hate scripting languages.

awscrt/http.py
```diff
@@ -71,12 +74,19 @@ def new(cls, bootstrap, host_name, port, socket_options,
         """
         future = Future()
         try:
-            assert isinstance(bootstrap, ClientBootstrap)
-            assert host_name
-            assert tls_connection_options is None or isinstance(tls_connection_options, TlsConnectionOptions)
+            assert isinstance(bootstrap, ClientBootstrap) or bootstrap is None
+            assert isinstance_str(host_name) and host_name
```
ColdenCullen: comment this nephew

awscrt/http.py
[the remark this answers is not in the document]
ColdenCullen: map isn't public tho?

elasticurl.py
```diff
@@ -81,7 +82,7 @@ def print_header_list(headers):
 # you only want one of these.
 event_loop_group = io.EventLoopGroup(1)
 
-# client bootstrap knows how to connect all the pieces. In this case it also has the default dns resolver
+ # client bootstrap knows how to connect all the pieces. In this case it also has the default dns resolver
```
ColdenCullen: wat

source/http_message.c
```diff
@@ -0,0 +1,238 @@
...
+    }
+
+    map = PyObject_GetAttrString(headers, "_map");
+    if (!map) {
+        goto done;
+    }
+
+    /* Clear existing headers, before adding the new values. */
+    size_t num_headers = aws_http_message_get_header_count(message->native);
+    for (size_t i = 0; i < num_headers; ++i) {
```
ColdenCullen:
```c
for (size_t i = num_headers; i > 0; --i) {
```

source/http_message.c
```diff
@@ -0,0 +1,238 @@
...
+        aws_http_message_erase_header(message->native, num_headers - (1 + i)); /* erase from back to front */
```
ColdenCullen:
```c
aws_http_message_erase_header(message->native, i - 1); /* erase from back to front */
```

source/http_message.c
```diff
@@ -0,0 +1,238 @@
...
+        }
+
+        if (!PyList_Check(values_list)) {
+            PyErr_SetString(PyExc_TypeError, "Header values should be in a list");
+        }
+
+        Py_ssize_t num_values = PyList_GET_SIZE(values_list);
+        for (Py_ssize_t value_i = 0; value_i < num_values; ++value_i) {
+            PyObject *name_value_pair = PyList_GET_ITEM(values_list, value_i);
+            PyObject *header_value = PyTuple_GetItem(name_value_pair, 1);
```
ColdenCullen: These 2 functions being named differently gave me cancer

source/io.c
```diff
@@ -609,3 +610,143 @@ PyObject *aws_py_tls_connection_options_set_server_name(PyObject *self, PyObject
...
+    PyGILState_STATE state = PyGILState_Ensure();
+
+    method_result = PyObject_CallMethod(impl->io, "seek", "(li)", &offset, &basis);
+    if (!method_result) {
+        aws_result = aws_py_raise_error();
+        goto done;
+    }
+
+    /* Clear EOF */
+    impl->is_end_of_stream = false;
```
ColdenCullen: What if someone seeks to END, 0? Should we check?

graebm: it would get set again the next time they tried to read.
This behavior seems consistent with [std::iostream](https://en.cppreference.com/w/cpp/io/ios_base/iostate)
> The following functions clear eofbit as a side-effect:
> -  basic_istream::putback
> -  basic_istream::unget
> -  basic_istream::seekg

source/http_message.c
graebm: The first one is the can't-fail-macro version, because I've already validated the length of the list I'm iterating.

The second one CAN fail, I'm checking something that I'm assuming is a tuple with 2 items.

I'll add comments

awscrt/http.py
graebm: Justin also thought this looked weird. Will comment

elasticurl.py
graebm: I removed these lines, then added them back a few commits later

source/http_message.c
graebm: that is much clearer. thanks

source/io.c
ColdenCullen: :+1:

PR #67, opened 2019-09-18
Track living NativeResources. Unit tests assert they're all gone.

graebm:

### DIFF
--- a/awscrt/__init__.py
+++ b/awscrt/__init__.py
@@ -12,6 +12,7 @@
 # permissions and limitations under the License.
 
 from sys import version_info
+from weakref import WeakSet
 
 __all__ = ['io', 'mqtt', 'crypto', 'http']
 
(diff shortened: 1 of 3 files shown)


awscrt/__init__.py
```diff
@@ -23,9 +24,22 @@ class NativeResource(object):
...
     """
+
+    track_lifetime = False
+    living = WeakSet()
+
     __slots__ = ('_binding', '__weakref__')
 
+    def __init__(self):
+        if NativeResource.track_lifetime:
+            NativeResource.living.add(self)
```
justinboswell: Pretty sure this is going to need to be protected due to cross-thread access, isn't it? Or is this all effectively under the GIL?

graebm: It's a `weakref.WeakSet`
I assumed it was magically thread-safe since objects can get cleaned up from any thread, but I'm not actually sure it is. Researching....

graebm: Ok, here's what my research shows:
- `WeakfSet` is [implemented in Python](https://github.com/python/cpython/blob/41c57b335330ff48af098d47e379e0f9ba09d233/Lib/_weakrefset.py#L35), not C.
- `WeakfSet.data` is the underlying `set` where things are stored.
- `set` is [implemented in C](https://github.com/python/cpython/blob/41c57b335330ff48af098d47e379e0f9ba09d233/Objects/setobject.c#L2038) so all its methods are atomic because GIL.
  - but stuff like looping over the set's members isn't thread-safe because iterating requires many distinct calls.
- I've found places where CPython uses `WeakfSet`. They don't seem to fear add/remove being thread-safe, but they will use locks or other [weird tricks](https://github.com/python/cpython/blob/a488879cbaf4b8b52699cadccf73bb4c271bcb29/Lib/asyncio/tasks.py#L45) if they're going to try and iterate it.

Conclusions:
- Multithreading won't corrupt `WeakSet`, but you might have problems if you 100% rely on the data. I do think it's OK for unit tests to assert that the WeakSet is empty.
- I don't want to keep this accessible, since it **might** be useful in a debugging scenario to iterate the members.

Therefore. I'm going to make this lifetime-tracking stuff `_private` and only access it from tests

justinboswell: :+1:

PR #68, opened 2019-09-19
ClientTlsContext tests and rename-o-rama

graebm:
The arguments to these functions confused me so I renamed them.
Will Captain Consistency come and kill me in my dreams now?


test/resources/unittests.crt
```diff
@@ -0,0 +1,23 @@
+-----BEGIN CERTIFICATE-----
```
justinboswell: Are these just random certs to see if the parser/IO works? They don't have any permissions anywhere right?

graebm:
> Does elasticurl.py need to be updated well?

nope. It was passing the arguments positionally

test/resources/unittests.crt
graebm: I copied them from aws-c-io's unit tests.
The unit tests I have so far are just testing the parser,
but when I get around to writing tests that make actual connections, we'll use them in similar ways to how they're used in aws-c-io

graebm: I assume they have no permissions, since they're publicly in aws-c-io

PR #75, opened 2019-09-24
New CI

graebm:
- Using new `builder.py` CI scheme
- Cleaned up `setup.py`
  - dependencies no longer built as a side-effect of import (they build as side-effect of the `build_ext` step)
  - get `/opt/openssl` on the docker images working by passing along `--library-dirs=...` and `--include-dirs=...` from the setup.py command line
- Upgrade `aws-c-mqtt` to v0.4.6


.travis.yml
```diff
@@ -13,11 +13,12 @@ addons:
...
-cache:
-  directories:
-    - ./packages
+install:
+  - export PATH=$(brew --prefix llvm)/bin:$PATH
 
 script:
-  - ./codebuild/common-macos.sh -DCMAKE_EXPORT_COMPILE_COMMANDS=ON
+  - export CODEBUILD_SRC_DIR=$(pwd)
+  - python3 -c "from urllib.request import urlretrieve; urlretrieve('https://raw.githubusercontent.com/awslabs/aws-c-common/master/codebuild/builder.py', 'builder.py')"
```
bretambrose: Would it make sense to tie this to the release/submodule of common  we're synced with?

graebm: that would make sense for aws-crt-python, but it would be inconsistent with how it's working in all the other repos.

setup.py
```diff
@@ -94,147 +100,143 @@ def determine_generator_string():
...
-        '--target', 'install',
-    ]
-
-    ret_code = subprocess.check_call(cmake_args, stderr=subprocess.STDOUT, shell=shell)
-    ret_code = subprocess.check_call(build_cmd, stderr=subprocess.STDOUT, shell=shell)
-
-    os.chdir(build_dir)
-    return ret_code
+        # continue with normal build_ext.run()
+        setuptools.command.build_ext.build_ext.run(self)  # can't use super() because base doesn't inherit from object
```
ColdenCullen: This comment is why I'm in therapy

PR #82, opened 2019-10-11
Resubscribe

graebm:
MQTT changes:
- added `mqtt.Connection.resubscribe_to_existing_topics()`
- `clean_session` flag is no longer ignored
- interrupted/resumed callbacks take additional `connection` arg
- Detect when server rejects the subscription

HTTP changes:
- `HttpStream.completion_future.result()` returns status_code instead of None 

Update dependencies:
- aws-c-common v0.4.12
- aws-c-io v0.4.12
- aws-c-mqtt to v0.4.9

### DIFF
--- a/awscrt/mqtt.py
+++ b/awscrt/mqtt.py
@@ -29,7 +29,7 @@ def _try_qos(qos_value):
     """Return None if the value cannot be converted to Qos (ex: 0x80 subscribe failure)"""
     try:
         return QoS(qos_value)
-    except BaseException:
+    except Exception:
         return None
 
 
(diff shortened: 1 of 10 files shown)


awscrt/mqtt.py
```diff
@@ -25,6 +25,14 @@ class QoS(IntEnum):
     EXACTLY_ONCE = 2
 
 
+def _try_qos(qos_value):
+    """Return None if the value cannot be converted to Qos (ex: 0x80 subscribe failure)"""
+    try:
+        return QoS(qos_value)
+    except BaseException:
```
ColdenCullen: From StackOverflow:
> You shouldn't catch BaseException - it includes SystemExit and KeyboardInterrupt, things you usually don't want to catch. 

Just catch regular Exception

source/mqtt_client_connection.c
```diff
@@ -423,7 +425,7 @@ PyObject *aws_py_mqtt_client_connection_connect(PyObject *self, PyObject *args)
                                                   .ping_timeout_ms = ping_timeout,
                                                   .on_connection_complete = s_on_connect,
                                                   .user_data = py_connection,
-                                                  .clean_session = true};
+                                                  .clean_session = is_clean_session};
```
ColdenCullen: While you're here, toss in a trailing comma to fix the formatting

PR #83, opened 2019-10-14
Simplify MSVC-picking logic in setup.py

graebm:
Python already has code that picks an MSVC version when it builds the extension. Use that.

Also update aws-c-common to v0.4.14 to fix Windows alignment issue.

I don't believe these things will fix the Windows-only crashes we're seeing in CI, but I needed these things to do local debugging.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -11,6 +11,7 @@
 # express or implied. See the License for the specific language governing
 # permissions and limitations under the License.
 
+from __future__ import print_function
 import distutils.ccompiler
 import glob
 import os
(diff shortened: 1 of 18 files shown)


setup.py
```diff
@@ -35,70 +35,49 @@ def is_arm():
...
-        print('Succesfully determined generator as \"{}\"'.format(vs_version_gen_str))
-        return vs_version_gen_str
-    return ''
+        try:
+            # See which compiler python picks
+            compiler = distutils.ccompiler.new_compiler()
+            compiler.initialize()
+
+            # Look at compiler path to divine the Visual Studio version
+            if '\\Microsoft Visual Studio\\2019' in compiler.cc:
```
JonathanHenson: NICE! and this will also work for the future.

bretambrose: Any worries about customers with non-standard install locations or do we assume that this is just controllable builds for publishing?

Are we assuming that this is being invoked from a dev-tools prompt or anywhere?

graebm: This works with vanilla cmd.exe or powershell,

You're right about it breaking with custom install paths. This is the best I could come up with without using private library functions. I added a comment in code explaining as much

PR #91, opened 2019-11-06
Actually bind Credentials to aws_credentials

graebm:
Previously, the `Credentials` class was pure python and I assumed we'd just copy data out as necessary.
When binding the signers I realized this was making things more complicated than if we'd just bound it directly to the `aws_credentials` struct.

### DIFF
--- a/awscrt/auth.py
+++ b/awscrt/auth.py
@@ -25,19 +25,12 @@ class Credentials(NativeResource):
 
     __slots__ = ()
 
-    @classmethod
-    def create(cls, access_key_id, secret_access_key, session_token=None):
+    def __init__(self, access_key_id, secret_access_key, session_token=None):
         assert isinstance_str(access_key_id)
         assert isinstance_str(secret_access_key)
         assert isinstance_str(session_token) or session_token is None
 
-        return cls._from_binding(_awscrt.credentials_new(access_key_id, secret_access_key, session_token))
-
-    @classmethod
-    def _from_binding(cls, credentials_binding):
-        credentials = cls()
-        credentials._binding = credentials_binding
-        return credentials
+        self._binding = _awscrt.credentials_new(access_key_id, secret_access_key, session_token)
 
     @property
     def access_key_id(self):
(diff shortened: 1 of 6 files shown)


awscrt/auth.py
```diff
@@ -18,21 +18,31 @@
...
 
 
-class Credentials(object):
+class Credentials(NativeResource):
     """
     Credentials are the public/private data needed to sign an authenticated AWS request.
     """
 
-    __slots__ = ('access_key_id', 'secret_access_key', 'session_token')
+    __slots__ = ()
```
ColdenCullen: uuuh how about `_binding`?

awscrt/auth.py
```diff
@@ -18,21 +18,31 @@
...
 
     def __init__(self, access_key_id, secret_access_key, session_token=None):
         assert isinstance_str(access_key_id)
         assert isinstance_str(secret_access_key)
         assert isinstance_str(session_token) or session_token is None
 
-        self.access_key_id = access_key_id
-        self.secret_access_key = secret_access_key
-        self.session_token = session_token
+        self._binding = _awscrt.credentials_new(access_key_id, secret_access_key, session_token)
```
ColdenCullen: should this be a call to super?

source/auth.h
```diff
@@ -16,6 +16,20 @@
  */
 #include "module.h"
 
+struct aws_credentials;
+
+PyObject *aws_py_credentials_new(PyObject *self, PyObject *args);
+
+/**
+ * Create Capsule to own pre-existing aws_credentials struct.
+ * The aws_credentials are destroyed when the capsule is destroyed.
```
ColdenCullen: this seems hecka dangerous. Why are we doing this?

awscrt/auth.py
graebm: It's in to `NativeResource.__slots__`
Slots are added by each step in the class hierarchy

awscrt/auth.py
graebm: you are correct

source/auth.h
graebm: oh. we're not anymore. I deleted this implementation

PR #92, opened 2019-11-07
AwsSigner & AwsSigningConfig

graebm:
- Everything in auth is prefixed with `Aws` because let's get real, none of this is generic.
  - Ok, "Signer" and "SigningConfig" could be interfaced, but I can introduce those concepts later when we need them. THIS IS PYTHOOOOOON
- `AwsSigningConfig` is immutable to discourage users from re-using it while it's still in use by an async signing operation.
  - totally sweet `.replace(...)` function clones the object, changing only the desired properties.


awscrt/auth.py
```diff
@@ -99,24 +129,212 @@ class DefaultCredentialsProviderChain(CredentialsProviderBase):
...
+        Returns a Future whose result will be the signed HttpRequest.
+
+        It is good practice to use a new config for each signature, or the date might get too old.
+        """
+        assert isinstance(http_request, HttpRequest)
+        assert isinstance(signing_config, AwsSigningConfig)
+
+        future = Future()
+
+        def _on_complete(error_code):
```
JonathanHenson: God I love python

source/auth_signer.c
```diff
@@ -0,0 +1,193 @@
...
+        goto done;
+    }
+
+    struct aws_http_message *http_request = aws_py_get_http_message(async_data->py_http_request);
+    if (!http_request) {
+        error_code = AWS_ERROR_UNKNOWN;
+        goto done;
+    }
+
+    if (aws_apply_signing_result_to_http_request(http_request, aws_py_get_allocator(), signing_result)) {
```
JonathanHenson: should we do this stuff prior to acquiring the GIL?

source/auth_signing_config.c
```diff
@@ -0,0 +1,292 @@
...
+}
+
+PyObject *aws_py_signing_config_get_credentials_provider(PyObject *self, PyObject *args) {
+    struct config_binding *binding = s_common_get(self, args);
+    if (!binding) {
+        return NULL;
+    }
+
+    Py_INCREF(binding->py_credentials_provider);
+    return binding->py_credentials_provider;
```
JonathanHenson: isn't it naturally going to be incremented when assigned in python land?

graebm: Returned values need their refcount bumped. Otherwise you'd never be able to return new objects from C to python. Ex: `return PyLong_FromLong(123);`

And it's why you can't just do `return Py_None;`. Instead you use the macro `Py_RETURN_NONE;` which does the refcount bump for you before returning.

Python must decref the returned value once it's in python land. If the user assigns it to a variable that's an INCREF and keeps it alive.

source/auth_signer.c
graebm: Yeah, we could do that

source/auth_signing_config.c
```diff
@@ -0,0 +1,292 @@
...
+/**
+ * Bind a Python AwsSigningConfig to a native aws_signing_config_aws.
+ */
+struct config_binding {
+    struct aws_signing_config_aws native;
+
+    struct aws_byte_buf string_storage;
+
+    /**
+     * Python objects that must must outlive this.
```
bretambrose: must x 2

test/test_auth.py
```diff
@@ -0,0 +1,251 @@
...
+SIGV4TEST_ACCESS_KEY_ID = 'AKIDEXAMPLE'
+SIGV4TEST_SECRET_ACCESS_KEY = 'wJalrXUtnFEMI/K7MDENG+bPxRfiCYEXAMPLEKEY'
+SIGV4TEST_SESSION_TOKEN = '6e86291e8372ff2a2260956d9b8aae1d763fbf315fa00fa31553b73ebf194267'
+SIGV4TEST_SERVICE = 'service'
+SIGV4TEST_REGION = 'us-east-1'
+SIGV4TEST_METHOD = 'GET'
+SIGV4TEST_PATH = '/'
+SIGV4TEST_UNSIGNED_HEADERS = [
+    ('Host', 'example.amazonaws.com'),
+    ('X-Amz-Date', '20150830T123600Z')]
```
ColdenCullen: del

test/test_auth.py
```diff
@@ -0,0 +1,251 @@
...
+class TestSigner(NativeResourceTest):
+
+    def test_create(self):
+        signer = awscrt.auth.AwsSigner()
+
+    def test_signing_sigv4_headers(self):
+        signer = awscrt.auth.AwsSigner()
+
+        credentials_provider = awscrt.auth.StaticAwsCredentialsProvider(
+            SIGV4TEST_ACCESS_KEY_ID, SIGV4TEST_SECRET_ACCESS_KEY, SIGV4TEST_SESSION_TOKEN)
```
ColdenCullen: Something's wrong here, because your session token isn't making it into the final header results

test/test_auth.py
graebm: fixed. passing date via signing config now

test/test_auth.py
graebm: WHOOPS. yes. shouldn't be using session key. This whole test was broken because I was comparing against the wrong thing. Fixed now

PR #93, opened 2019-11-14
Bind aws_http_headers to awscr.http.HttpHeaders

graebm:
Also added awscrt.io.InputSteam class to bind to aws_input_stream.

Also bound HttpMessage more tightly to aws_http_message


awscrt/io.py
```diff
@@ -281,3 +283,30 @@ def set_alpn_list(self, alpn_list):
 
     def set_server_name(self, server_name):
         _awscrt.tls_connection_options_set_server_name(self, server_name)
+
+
+class InputStream(NativeResource):
```
JonathanHenson: consideration, support for with .... autoclose, whatever it's called.

graebm: This class take a python `IOBase` instance in its constructor. It would be up to whoever opened *that* instance to close that instance.

PR #94, opened 2019-11-15
Subscribe to any + MQTT tests

justinboswell:


awscrt/mqtt.py
```diff
@@ -207,14 +205,26 @@ def suback(packet_id, topic, qos, error_code):
...
-            assert callable(callback)
+            assert callable(callback) or callback is None
             assert isinstance(qos, QoS)
             packet_id = _awscrt.mqtt_client_connection_subscribe(self._binding, topic, qos.value, callback, suback)
         except Exception as e:
             future.set_exception(e)
 
         return future, packet_id
 
+    def subscribe_to_any(self, callback):
```
ColdenCullen: I don't like this name because it implies that we're actually doing a SUBSCRIBE operation. How about on_any_publish or something? You could make it a set only property to make it clearer.

source/mqtt_client_connection.c
```diff
@@ -692,14 +702,42 @@ PyObject *aws_py_mqtt_client_connection_subscribe(PyObject *self, PyObject *args
         suback_callback);
 
     if (msg_id == 0) {
-        Py_DECREF(callback);
-        Py_DECREF(suback_callback);
+        if (callback) {
+            Py_DECREF(callback);
```
ColdenCullen: These can just be XDECREFs

justinboswell:
The windows failures are the same ones @graebm was seeing. Everything else works just fine, and those tests pass on my windows machine.

source/mqtt_client_connection.c
```diff
@@ -643,6 +644,10 @@ static void s_suback_callback(
     (void)connection;
 
     PyObject *callback = userdata;
+    if (!callback) {
+        return;
```
graebm: The way your code is now, you want to either:
- check `if(callback != Py_None)`
- OR when you're passing the userdata to C, do `(callback == Py_None) ? NULL : callback`

I personally prefer the 1st approach, but it's apples and oranges

source/mqtt_client_connection.c
```diff
@@ -692,14 +698,38 @@ PyObject *aws_py_mqtt_client_connection_subscribe(PyObject *self, PyObject *args
         suback_callback);
 
     if (msg_id == 0) {
-        Py_DECREF(callback);
-        Py_DECREF(suback_callback);
+        Py_XDECREF(callback);
+        Py_XDECREF(suback_callback);
```
graebm: likewise, can't be NULL

source/mqtt_client_connection.c
```diff
@@ -692,14 +698,38 @@ PyObject *aws_py_mqtt_client_connection_subscribe(PyObject *self, PyObject *args
...
+        return NULL;
+    }
+
+    struct mqtt_connection_binding *py_connection =
+        PyCapsule_GetPointer(impl_capsule, s_capsule_name_mqtt_client_connection);
+    if (!py_connection) {
+        return NULL;
+    }
+
+    callback = PyWeakref_NewProxy(callback, NULL);
```
graebm: 2 things:
1) not checking for NULL return
2) if this succeeds, we never DECREF the callback, so the proxy will leak. You probably want to store a member on the binding object `PyObject *on_message;` so we can DECREF it in the binding destructor.
    - Also, in this fn, be sure to protect against someone setting on_message multiple times. DECREF the existing value before setting the new callback and INCREFing it

test/test_mqtt.py
```diff
@@ -0,0 +1,132 @@
...
+            self.valid = False
+
+    @staticmethod
+    def get():
+        if Config.cache:
+            return Config.cache
+
+        # boto3 caches the HTTPS connection for the API calls, which appears to the unit test
+        # framework as a leak, so ignore it, that's not what we're testing here
+        warnings.simplefilter('ignore', ResourceWarning)
```
graebm: trival typo: ...**has** a leak...

justinboswell: Nope, this one is correct :)

PR #99, opened 2019-11-21
Proxy & MQTT over websockets

graebm:
- Proxy has testing.
- Websockets does not have it yet, it will be added in the next PR along with the EZ-IOT-BUILDER.
- Made MQTT tests Python 2 compatible
- Also changed how we solved the MQTT weakref problem. In https://github.com/awslabs/aws-crt-python/pull/94 Justin stored a weakref to each callback in the binding. Before that PR was up, I'd solved the same issue in this branch differently, keeping 1 weakref to the Connection itself in the binding, and invoking callbacks by name. I was merging Justin's solution over mine, but ran into a snag when I needed to make a strong-ref from the weak-ref during the websocket handshake. At that point it became easier to merge my stuff over Justin's.


awscrt/mqtt.py
```diff
@@ -85,31 +87,100 @@ def __init__(self,
...
+            _awscrt.mqtt_ws_handshake_transform_complete(f.exception(), native_userdata)
+
+        future = Future()
+        future.add_done_callback(_on_complete)
+        http_request = HttpRequest._from_bindings(http_request_binding, http_headers_binding)
+        transform_args = WebsocketHandshakeTransformArgs(self, http_request, future)
+        try:
+            self._ws_handshake_transform_cb(transform_args)
+        except Exception as e:
+            traceback.print_exc()
```
justinboswell: Don't all exceptions capture a stack? What's the impetus here?

source/mqtt_client_connection.c
```diff
@@ -95,17 +100,17 @@ static void s_on_connection_interrupted(struct aws_mqtt_client_connection *conne
...
     PyGILState_STATE state = PyGILState_Ensure();
 
-    PyObject *result = PyObject_CallFunction(py_connection->on_connection_interrupted, "(i)", error_code);
-    if (result) {
-        Py_DECREF(result);
-    } else {
-        PyErr_WriteUnraisable(PyErr_Occurred());
+    /* Ensure that python class is still alive */
+    PyObject *self = PyWeakref_GetObject(py_connection->self_proxy); /* borrowed reference */
+    if (self != Py_None) {
```
justinboswell: Oh I like this much better

awscrt/mqtt.py
graebm: I was trying to give the user more information, the Python exception will be swallowed as we translate this to an AWS_ERROR_HTTP_CALLBACK_FAILURE that stops the MQTT connection attempt. So I was trying to drop a breadcrumb.

But this isn't the best place to dump that information.

PR #104, opened 2019-11-25
AwsCrtError class instead of error_code ints

graebm:
- Example: `AwsCrtError(name='AWS_IO_SOCKET_CLOSED', message='socket is closed.', code=1051)`
- Remap to built-in python exceptions where possible (Ex: Return `MemoryError` instead of `AwsCrtError(name='AWS_ERROR_OOM')`
- In the future, I still intend to code-gen a class for each AWS_ERROR int, but this is a forwards-compatible stand-in until then (generated classes will all inherit from AwsCrtError).

### DIFF
--- a/awscrt/http.py
+++ b/awscrt/http.py
@@ -23,32 +23,26 @@ from enum import IntEnum
 class HttpConnectionBase(NativeResource):
     """
     Base for HTTP connection classes.
+
+    Attributes:
+        shutdown_future (concurrent.futures.Future): Completes when the connection has finished shutting down.
+                Future will contain a result of None, or an exception indicating why shutdown occurred.
+                Note that the connection may have been garbage-collected before this future completes.
     """
 
-    __slots__ = ('_shutdown_future')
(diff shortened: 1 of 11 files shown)


awscrt/http.py
```diff
@@ -102,12 +105,20 @@ def on_connection_setup(binding, error_code):
...
-                    future.set_exception(Exception("Error during connect: err={}".format(error_code)))
+                    future.set_exception(awscrt.exceptions.from_code(error_code))
+
+            # on_shutdown MUST NOT reference the connection itself, just the _shutdown_future within it.
+            # Otherwise we create a circular reference that prevents the connection from getting GC'd.
+            shutdown_future = connection._shutdown_future
+
+            def on_shutdown(error_code):
+                error = None if error_code == 0 else awscrt.exceptions.from_code(error_code)
+                shutdown_future.set_result(error)
```
ColdenCullen: raise this bad boi

graebm: done

PR #109, opened 2019-12-02
Callbacks take kwargs in the suggested manner

graebm:
No API changes here, just fixing up existing callbacks. 

- Add "forward compatibility kwargs" to callbacks that were missing it.
- Put back named args in functions that were **only** taking `(**kwargs)`. This is much nicer to read and write. This is what they Python SDK team had meant, but I didn't totally understand on the first pass.
  - Ex: ~~def my_msg_cb(**kwargs)~~ -> `def my_msg_cb(topic, message, **kwargs)`

### DIFF
--- a/mqtt_test.py
+++ b/mqtt_test.py
@@ -34,10 +34,10 @@ parser.add_argument('--root-ca', help="File path to root certificate authority, 
 
 io.init_logging(LogLevel.Trace, 'stderr')
 
-def on_connection_interrupted(connection, error):
+def on_connection_interrupted(connection, error, **kwargs):
     print("Connection has been interrupted with error", error)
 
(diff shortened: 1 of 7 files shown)


mqtt_test.py
```diff
@@ -59,8 +59,8 @@ def on_resubscribe_complete(resubscribe_future):
 
 receive_results = {}
 receive_event = threading.Event()
-def on_receive_message(**kwargs):
-    receive_results.update(kwargs)
+def on_receive_message(topic, payload, **kwargs):
+    receive_results.update(locals())
```
ColdenCullen: I'd suggest reverting this one, since you're not doing anything with the named variables anyway...

graebm: I made this more explicit and less weird

PR #116, opened 2019-12-18
Support sloppy shutdowns

graebm:
Add tests that ensure we can shutdown cleanly with native stuff still running. Added lots of stuff to make this work and get our CI more robust.

Windows CI is now required! 🎉🎉🎉 It always passes! 🎉🎉🎉

- test_appexit.py: runs sample apps that exit with native stuff still running. Checks that exit-code is zero.
  - appexit_http.py: sample app that makes an HTTP request
  - TODO: mqtt sample app
- Expose EventLoopGroup shutdown to python.
- Don't acquire GIL during application shutdown. Attempting to do so will crash.
- Python 2 runs shutdown fn. It didn't used to, and the logger would raises errors during shutdown.
- Don't AWS_FATAL_ASSERT that python callbacks succeed. They often fail during shutdown.
- Increase connection timeouts to 5s by default. elasticurl and mqtt-builder used 3s, which often failed on Windows CI.
- Add/use helper functions for reading basic types from PyObject attributes (ex: `PyObject_GetAttrAsUint16()`). Fixed a bug, simplified a bunch of code, and makes error reporting more robust.

### DIFF
--- a/source/module.c
+++ b/source/module.c
@@ -596,10 +596,10 @@ PyMODINIT_FUNC INIT_FN(void) {
         s_module_doc,
         -1, /* size of per-interpreter state of the module, or -1 if the module keeps state in global variables. */
         s_module_methods,
-        NULL,
-        NULL,
-        NULL,
-        s_module_free,
+        NULL, /* slots for multi-phase initialization */
+        NULL, /* traversal fn to call during GC traversal */
+        NULL, /* clear fn to call during GC clear */
+        s_module_free, /* fn to call during deallocation of the module object */
     };
     PyObject *m = PyModule_Create(&s_module_def);
 #elif PY_MAJOR_VERSION == 2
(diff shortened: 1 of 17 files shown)


source/module.c
```diff
@@ -489,41 +587,64 @@ static void s_module_free(void *userdata) {
...
 
-#if PY_MAJOR_VERSION == 3
-    return m;
-#endif /* PY_MAJOR_VERSION */
+    s_module_init();
 }
+
+#else
+#    error Unsupported Python version
+#endif /* PY_MAJOR_VERSION */
```
JonathanHenson: newline

test/appexit_http.py
```diff
@@ -0,0 +1,156 @@
...
+
+
+def set_stage(stage):
+    print("Stage:", stage)
+    if stage.value >= EXIT_STAGE.value:
+        print("Exiting now")
+        sys.exit()
+
+
+if __name__ == '__main__':
```
JonathanHenson: This is awesome!!!!

Issue #128, opened 2020-02-16
Inconsistent keep alive values


burritoIand:
Hello,

The default keep_alive value for the mqtt client is 3600 seconds.
https://github.com/awslabs/aws-crt-python/blob/master/awscrt/awsiot_mqtt_connection_builder.py#L185
https://github.com/awslabs/aws-crt-python/blob/master/awscrt/mqtt.py#L80

In the pub/sub example from v2 sdk it is 6:
https://github.com/aws/aws-iot-device-sdk-python-v2/blob/master/samples/pubsub.py#L120

When experimenting I discovered that if I use the default and don't set my own value, I receive an error log from AWS IoT:
```
{
    "timestamp": "2020-02-16 21:08:56.976",
    "logLevel": "ERROR",
    "status": "Failure",
    "eventType": "Unsupported keep-alive value",
    "protocol": "MQTT",
    "clientId": "API::0188036e-dce3-4bdb-9e3b-dcff46c448b8",
    "principalId": "AROAU2IT4DXC5K2MPHPLL:foo-AppEntryFunction-1M4ADAE3QQEOR",
    "details": "An unsupported MQTT keep-alive value was requested from the server. AWS IoT supports keep-alive values between 5 and 1200. The connection will be handled with the supported value closest to the request value."
}
```
Is the value of 3600 still accurate and good for default?

graebm:
Where are you getting that error log from? I have not seen that error before, and don't even know where to look for errors like that!

I think you're right, the SDK default values are too high. [This doc](https://docs.aws.amazon.com/iot/latest/developerguide/limits-iot.html) says:
> The default keep-alive interval is 1200 seconds. If a client requests a keep-alive interval of zero, the default keep-alive interval is used. If a client requests a keep-alive interval longer than 1200 seconds, the default keep-alive interval is used. If a client requests a keep-alive interval shorter than 30 seconds but longer than zero, the server treats the client as though it requested a keep-alive interval of 30 seconds.

So that doc and your error log, agree that 1200 seconds is the max (though they differ slightly in what the minimum value is).

You should set a lower keep-alive, and I'll work on fixing this in the SDK

burritoIand:
Hey @graebm ok great thanks for looking into it! That document you linked has some good stuff, I hadn't read some of the limits on things etc.

graebm:
Alright fixes published to pip:
awscrt 0.5.11
and
[...]

PR #130, opened 2020-02-18
Update Dependencies. Misc improvements

graebm:
- Update setup.py version to 0.5.7
- Update dependencies
- Small improvements
  - Reduce default MQTT keep-alive values (see Issue https://github.com/awslabs/aws-crt-python/issues/128)
  - Simplify README
  - Better error message if `cmake` not installed (see [forum issue](https://forums.aws.amazon.com/thread.jspa?threadID=316634&tstart=0))
- CI improvements
  - Set up Github secrets so MQTT tests will run again
  - Alter credentials tests so they can't accidentally leak secrets
  - `ManyLinuxCI` action uses `AWSCrtPython` action.


awscrt/awsiot_mqtt_connection_builder.py
```diff
@@ -182,7 +182,7 @@ def _builder(
         clean_session=kwargs.get('clean_session', False),
         reconnect_min_timeout_secs=kwargs.get('reconnect_min_timeout_secs', 5),
         reconnect_max_timeout_secs=kwargs.get('reconnect_max_timeout_secs', 60),
-        keep_alive_secs=kwargs.get('keep_alive_secs', 3600),
+        keep_alive_secs=kwargs.get('keep_alive_secs', 1200),
```
justinboswell: Docs say this value must be higher than ping_timeout_ms, but that is not the case in the defaults.

graebm: One value is in seconds, the other in milliseconds. Changed the wording of the comments slightly to try and make this less confusing
~~This value must be higher than ping_timeout_ms.~~
This duration must be longer than ping_timeout_ms.

PR #136, opened 2020-03-17
apply http stream activation refactor.

JonathanHenson:


source/http_stream.c
```diff
@@ -327,3 +327,22 @@ PyObject *aws_py_http_client_stream_new(PyObject *self, PyObject *args) {
...
     return NULL;
 }
+
+PyObject *aws_py_http_client_stream_activate(PyObject *self, PyObject *args) {
+    (void)self;
+
+    PyObject *py_stream = NULL;
+    if (!PyArg_ParseTuple(args, "O", &py_stream)) {
+        PyErr_SetString(PyExc_TypeError, "stream object must not be null");
+        Py_RETURN_FALSE;
```
graebm: Just `return NULL;`

for a few reasons:
- If PyArg_ParseTuple() fails, it's already set a helpful python exception explaining why it failed.
- PyArg_ParseTuple() will accept `None` for "O" (object) types, so this error message isn't even accurate

source/http_stream.c
```diff
@@ -327,3 +327,22 @@ PyObject *aws_py_http_client_stream_new(PyObject *self, PyObject *args) {
...
+    }
+
+    struct aws_http_stream *native_stream = aws_py_get_http_stream(py_stream);
+
+    if (aws_http_stream_activate(native_stream)) {
+        PyErr_AwsLastError();
+        Py_RETURN_FALSE;
```
graebm:
```
return PyErr_AwsLastError();
```

this function is a shortcut for `PyErr_SetAwsLastError(); return NULL;`
and NULL needs to be returned to trigger an exception

source/http_stream.c
```diff
@@ -327,3 +327,22 @@ PyObject *aws_py_http_client_stream_new(PyObject *self, PyObject *args) {
...
+    }
+
+    Py_RETURN_TRUE;
```
graebm: Py_RETURN_NONE;

don't see why we should return true/false from this function

source/http_stream.c
graebm: fixed

source/http_stream.c
graebm: fixed

source/http_stream.c
graebm: fixed

PR #143, opened 2020-05-01
Fix bug in binding of aws_input_stream_seek()

graebm:
Testing this bug via an auth test, rather than a test direcly on `awscrt.io.InputStream` because we haven't added read()/seek() functions that are accessible from python. But we know that auth will seek after signing the body

Don't have a canonical Sigv4 results to check the signed request against, so this new test just checks that we don't raise an exception. Can anyone hook me up with expected signed values?

### DIFF
--- a/test/test_auth.py
+++ b/test/test_auth.py
@@ -286,4 +286,10 @@ class TestSigner(NativeResourceTest):
 
         self.assertIs(http_request, signing_result)  # should be same object
 
-        self.assertEqual(0, body_stream.tell())  # stream's position should be at beginning
+        # don't have canonical headers to test against
+        # so just assert that these expected headers are present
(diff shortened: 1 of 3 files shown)


test/test_auth.py
```diff
@@ -257,3 +258,32 @@ def test_signing_sigv4_headers(self):
...
+        http_request = awscrt.http.HttpRequest(
+            method='POST',
+            path='/',
+            headers=awscrt.http.HttpHeaders([('Host', 'example.amazonaws.com'), ('Content-Length', '5')]),
+            body_stream=body_stream)
+
+        signing_future = awscrt.auth.aws_sign_request(http_request, signing_config)
+
+        signing_result = signing_future.result(TIMEOUT)
+
```
JonathanHenson, line 286: So we want to assert that the signed header is now there?

PR #158, opened 2020-05-29
Added python script to do version update, called from all platform builds

justinboswell:

### DIFF
--- a/continuous-delivery/update-version.py
+++ b/continuous-delivery/update-version.py
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python3
 
 import os
 import re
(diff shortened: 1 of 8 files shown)


continuous-delivery/update-version.py
```diff
@@ -0,0 +1,16 @@
+#!/usr/bin/env python
```
graebm, line 1: should this be python3?

https://stackoverflow.com/questions/6908143/should-i-put-shebang-in-python-scripts-and-what-form-should-it-take

continuous-delivery/update-version.py
```diff
@@ -0,0 +1,16 @@
...
+import re
+import subprocess
+
+tag = subprocess.check_output(['git', 'describe', '--tags'])
+# strip the leading v
+version = str(tag[1:].strip(), 'utf8')
+setup_path = os.path.join(os.path.dirname(__file__), '..', 'setup.py')
+print("Updating setup.py to version {}".format(version))
+with open(setup_path, 'r+') as setup_py:
+    contents = setup_py.read(64 * 1024)
```
graebm, line 13: It's a bit scary to pass in a "this should always be big enough" value

I think read() with no arguments will read the whole file
or maybe there's a readall() function?
I have a hard time determining the exact API with the ambiguous [io](https://docs.python.org/3/library/io.html) API.

continuous-delivery/update-version.py
graebm: actually, the whole line seems unnecessary

PR #167, opened 2020-07-07
Surface publish and unsubscribe errors back to Python

bretambrose:
https://github.com/aws/aws-iot-device-sdk-python-v2/issues/62

### DIFF
--- a/awscrt/mqtt.py
+++ b/awscrt/mqtt.py
@@ -543,10 +543,11 @@ class Connection(NativeResource):
         future = Future()
         packet_id = 0
 
-        def unsuback(packet_id):
-            future.set_result(dict(
-                packet_id=packet_id
-            ))
+        def unsuback(packet_id, error_code):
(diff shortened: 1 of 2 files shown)


awscrt/mqtt.py
```diff
@@ -543,10 +543,11 @@ def unsubscribe(self, topic):
         future = Future()
         packet_id = 0
 
-        def unsuback(packet_id):
-            future.set_result(dict(
-                packet_id=packet_id
-            ))
+        def unsuback(packet_id, error_code):
```
JonathanHenson, line 546: is this going to cause backwards compatibility problems?

bretambrose: These are private callbacks within the implementation, so they shouldn't.

graebm: Yep, this is fine,
it's just a private function that routes info from the CRT to the actual `future` the user is dealing with

PR #179, opened 2020-09-23
doc: Highlight that the retain option is not supported by AWS IoT

JoshM1994:
* Add a link to the AWS IoT documentation specifying that the retain flag will cause a disconnection
* https://github.com/aws/aws-iot-device-sdk-python-v2/issues/113\#issuecomment-697995436

*Issue #, if available:*

*Description of changes:*

Add a link to reference documentation highlighting the the `retain` flag is not supported by AWS IoT

### DIFF
--- a/awscrt/mqtt.py
+++ b/awscrt/mqtt.py
@@ -110,12 +110,14 @@ class Will:
         qos (QoS): QoS used when publishing the Will message.
         payload (bytes): Content of Will message.
         retain (bool): Whether the Will message is to be retained when it is published.
+            This is not supported by AWS IoT. See https://docs.aws.amazon.com/iot/latest/developerguide/mqtt.html
 
     Attributes:
         topic (str): Topic to publish Will message on.
         qos (QoS): QoS used when publishing the Will message.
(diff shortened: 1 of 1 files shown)


awscrt/mqtt.py
```diff
@@ -110,12 +110,14 @@ class Will:
         qos (QoS): QoS used when publishing the Will message.
         payload (bytes): Content of Will message.
         retain (bool): Whether the Will message is to be retained when it is published.
+            This is not supported by AWS IoT. See https://docs.aws.amazon.com/iot/latest/developerguide/mqtt.html
```
graebm, line 113: Thanks, this is good to call out, but let's keep it ambiguous in case this feature is added in the future and we forget to update all the docs everywhere. I know the AWS IoT team is always weighing the possibility of supporting more MQTT features.

```Retain is not supported by AWS IoT as of 2020, see https://docs.aws.amazon.com/iot/latest/developerguide/mqtt.html for current status.```

bretambrose:
Given the recent launch with retain support, this is no longer applicable.

PR #184, opened 2020-10-06
Python 3.9 updates

justinboswell:

### DIFF
--- a/.builder/actions/manylinux-ci.py
+++ b/.builder/actions/manylinux-ci.py
@@ -45,7 +45,7 @@ class ManyLinuxPackage(Builder.Action):
             steps.append(Builder.Script(actions, name=python))
 
         copy_steps = [
-            ['cp', '-r', 'wheelhouse', '../dist']
+            ['cp', '-r', 'wheelhouse', '../dist'],
             ['cp', 'dist/*.tar.gz', '../dist/']
         ]
 
--- a/.builder/actions/manylinux-ci.py
+++ b/.builder/actions/manylinux-ci.py
@@ -5,7 +5,6 @@ import sys
 from aws_crt_python import AWSCrtPython, InstallPythonReqs
 
 pythons = (
-    'cp35-cp35m',
     'cp36-cp36m',
     'cp37-cp37m',
     'cp38-cp38m',
--- a/.builder/actions/manylinux-ci.py
+++ b/.builder/actions/manylinux-ci.py
@@ -5,10 +5,11 @@ import sys
 from aws_crt_python import AWSCrtPython, InstallPythonReqs
 
 pythons = (
+    'cp35-cp35m',
     'cp36-cp36m',
     'cp37-cp37m',
-    'cp38-cp38m',
-    'cp39-cp39m',
+    'cp38-cp38',
+    'cp39-cp39',
 )
 
 
--- a/.builder/actions/manylinux-ci.py
+++ b/.builder/actions/manylinux-ci.py
@@ -8,6 +8,8 @@ pythons = (
     'cp35-cp35m',
     'cp36-cp36m',
     'cp37-cp37m',
+    'cp38-cp38m',
+    'cp39-cp39m',
 )
 
 
--- a/.builder/actions/manylinux-ci.py
+++ b/.builder/actions/manylinux-ci.py
@@ -45,7 +45,7 @@ class ManyLinuxPackage(Builder.Action):
             steps.append(Builder.Script(actions, name=python))
 
         copy_steps = [
-            ['cp', '-r', 'wheelhouse' '../dist']
+            ['cp', '-r', 'wheelhouse', '../dist']
             ['cp', 'dist/*.tar.gz', '../dist/']
         ]
 
--- a/setup.py
+++ b/setup.py
@@ -110,6 +110,7 @@ PROJECT_DIR = os.path.dirname(os.path.realpath(__file__))
 DEP_BUILD_DIR = os.path.join(PROJECT_DIR, 'build', 'deps')
 DEP_INSTALL_PATH = os.environ.get('AWS_C_INSTALL', os.path.join(DEP_BUILD_DIR, 'install'))
 
+AWS_LIBCRYPTO_INSTALL = None
 if sys.platform != 'darwin' and sys.platform != 'win32':
     AWS_LIBCRYPTO_INSTALL = os.environ.get('AWS_LIBCRYPTO_INSTALL', os.path.join(DEP_BUILD_DIR, 'libcrypto'))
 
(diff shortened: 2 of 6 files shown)


.builder/actions/manylinux-ci.py
```diff
@@ -43,7 +45,7 @@ def run(self, env):
             steps.append(Builder.Script(actions, name=python))
 
         copy_steps = [
-            ['cp', '-r', 'wheelhouse' '../dist']
+            ['cp', '-r', 'wheelhouse', '../dist'],
```
graebm, line 48: O_O
how did this ever work?

setup.py
```diff
@@ -110,6 +110,9 @@ def __init__(self, name, extra_cmake_args=[]):
 DEP_BUILD_DIR = os.path.join(PROJECT_DIR, 'build', 'deps')
 DEP_INSTALL_PATH = os.environ.get('AWS_C_INSTALL', os.path.join(DEP_BUILD_DIR, 'install'))
 
+if sys.platform != 'darwin' and sys.platform != 'win32':
+    AWS_LC_INSTALL = os.environ.get('AWS_LC_INSTALL', os.path.join(DEP_BUILD_DIR, 'libcrypto'))
```
graebm, line 114: is `AWS_LC_INSTALL` a standard environment variable we're using in other CRTs as well?

If it's something we're introducing here for the 1st time, let's remove the acronym and name it `AWS_LIBCRYPTO_INSTALL`

setup.py
```diff
@@ -142,15 +145,19 @@ def _build_dependency(self, aws_lib):
...
         if self.include_dirs:
-            cmake_args.append('-DCMAKE_INCLUDE_PATH={}'.format(';'.join(self.include_dirs)))
+            cmake_args.append('-DCMAKE_INCLUDE_PATH="{}"'.format(';'.join(self.include_dirs)))
         if self.library_dirs:
-            cmake_args.append('-DCMAKE_LIBRARY_PATH={}'.format(';'.join(self.library_dirs)))
+            cmake_args.append('-DCMAKE_LIBRARY_PATH="{}"'.format(';'.join(self.library_dirs)))
+        if sys.platform != 'darwin' and sys.platform != 'win32':
+            cmake_args.append('-DLibCrypto_INCLUDE_DIR={}/include'.format(AWS_LC_INSTALL))
+            cmake_args.append('-DLibCrypto_STATIC_LIBRARY={}/lib/libcrypto.a'.format(AWS_LC_INSTALL))
+            self.library_dirs.append('{}/lib'.format(AWS_LC_INSTALL))
```
graebm, line 156: what about `lib64`?
😐

setup.py
```diff
@@ -160,6 +167,7 @@ def _build_dependency(self, aws_lib):
             '--config', build_type,
             '--target', 'install',
         ]
+        print(' '.join(build_cmd))
```
graebm, line 170: `print(subprocess.list2cmdline(build_cmd))`

setup.py
```diff
@@ -142,15 +145,19 @@ def _build_dependency(self, aws_lib):
             '-DBUILD_SHARED_LIBS=OFF',
             '-DCMAKE_BUILD_TYPE={}'.format(build_type),
             '-DBUILD_TESTING=OFF',
-            '-DS2N_NO_PQ_ASM=ON',
         ])
         if self.include_dirs:
-            cmake_args.append('-DCMAKE_INCLUDE_PATH={}'.format(';'.join(self.include_dirs)))
+            cmake_args.append('-DCMAKE_INCLUDE_PATH="{}"'.format(';'.join(self.include_dirs)))
```
graebm, line 150: I've wondered what the right way to quote these is...

subprocess.list2cmdline() will turn this into: `cmake "-DCMAKE_INCLUDE_PATHS=\"my crappy path\"" ...`

Should there be the escaped quotes, within the other quotes, like that?

setup.py
justinboswell: ok

.builder/actions/manylinux-ci.py
justinboswell: NO IDEA

setup.py
justinboswell: libcrypto doesn't use lib64 conventions

setup.py
justinboswell: Either way works, I prefer to avoid the escapes in the stuff we have to read, python fixes the quoting before passing it to the shell

graebm: Yeah, I meant, do we need to add quotes at all? because python will add quotes around the whole argument in the end

setup.py
```diff
@@ -110,6 +110,9 @@ def __init__(self, name, extra_cmake_args=[]):
 DEP_BUILD_DIR = os.path.join(PROJECT_DIR, 'build', 'deps')
 DEP_INSTALL_PATH = os.environ.get('AWS_C_INSTALL', os.path.join(DEP_BUILD_DIR, 'install'))
 
+if sys.platform != 'darwin' and sys.platform != 'win32':
+    AWS_LIBCRYPTO_INSTALL = os.environ.get('AWS_LIBCRYPTO_INSTALL', os.path.join(DEP_BUILD_DIR, 'libcrypto'))
```
graebm, line 114: trivial/style: let's set `AWS_LIBCRTYPO_INSTALL = None` in an else-statement, or before the if-statement.

then later we can just `if AWS_LIBCRYPTO_INSTALL:` instead of repeating this complex if-statement again

setup.py
```diff
@@ -142,15 +145,19 @@ def _build_dependency(self, aws_lib):
...
-            '-DS2N_NO_PQ_ASM=ON',
         ])
         if self.include_dirs:
-            cmake_args.append('-DCMAKE_INCLUDE_PATH={}'.format(';'.join(self.include_dirs)))
+            cmake_args.append('-DCMAKE_INCLUDE_PATH="{}"'.format(';'.join(self.include_dirs)))
         if self.library_dirs:
-            cmake_args.append('-DCMAKE_LIBRARY_PATH={}'.format(';'.join(self.library_dirs)))
+            cmake_args.append('-DCMAKE_LIBRARY_PATH="{}"'.format(';'.join(self.library_dirs)))
+        if sys.platform != 'darwin' and sys.platform != 'win32':
+            cmake_args.append('-DLibCrypto_INCLUDE_DIR={}/include'.format(AWS_LIBCRYPTO_INSTALL))
```
graebm, line 154: trivial/style: related to other comment, let's simplify this to `if AWS_LIBCRYPTO_INSTALL:`

graebm (approved):

setup.py
justinboswell: Good call!

PR #194, opened 2020-10-27
Actually use manylinux2014 docker image.

graebm:
Actually use the manylinux2014 docker images, instead of just running on an EC2 instance that's similar enough that we happen to pass the wheel audit.

Also, setup.py improvements:
- accept `cmake3` as well as `cmake`.
- slight cleanup of how AWS_LIBCRYPTO_INSTALL is used.


setup.py
```diff
@@ -73,17 +74,54 @@ def determine_generator_args():
...
+
+    for cmake_alias in ['cmake3', 'cmake']:
+        cmake_found = shutil.which(cmake_alias)
+        if cmake_found:
+            return cmake_found
+
+    raise Exception("CMake must be installed to build from source.")
+
+
+def get_libcrypto_static_library(libcrypto_dir):
```
justinboswell, line 93: Someone went in for extra credit

justinboswell (approved):

PR #201, opened 2020-10-31
event-stream RPC connection

graebm:
implemented connect() and close() so far. Chose to use a handler, instead of individual callbacks, so that it's harder to create circular references that would prevent the connection from ever cleaning up.

### DIFF
--- a/awscrt/eventstream.py
+++ b/awscrt/eventstream.py
@@ -29,12 +29,12 @@ class EventStreamRpcClientConnectionHandler:
     resources finish shutting down.
 
     Attributes:
-        connection: Initially None. Will be set to the actual connection
-            before any events are invoked.
+        connection (Optional[EventStreamRpcClientConnection]): Initially None.
+            Will be set to the actual connection before any events are invoked.
     """
 
     def __init__(self):
-        self.connection: Optional[EventStreamRpcClientConnection] = None
+        self.connection = None
 
     def on_connection_setup(self, **kwargs) -> None:
         """Invoked when connection has been successfully established.
@@ -66,7 +66,7 @@ class EventStreamRpcClientConnection(NativeResource):
 
         port (int): Remote port.
 
-        shutdown_future (concurrent.futures.Future): Completes when this
+        shutdown_future (concurrent.futures.Future[None]): Completes when this
             connection has finished shutting down. Future will contain a
             result of None, or an exception indicating why shutdown occurred.
             Note that the connection may have been garbage-collected before
@@ -78,9 +78,9 @@ class EventStreamRpcClientConnection(NativeResource):
     def __init__(self, host_name, port):
         # Do no instantiate directly, use static connect method
         super().__init__()
-        self.host_name: str = host_name
-        self.port: int = port
-        self.shutdown_future: Future[None] = Future()
+        self.host_name = host_name
+        self.port = port
+        self.shutdown_future = Future()
 
     @classmethod
     def connect(
--- a/awscrt/eventstream.py
+++ b/awscrt/eventstream.py
@@ -183,4 +183,10 @@ class EventStreamRpcClientConnection(NativeResource):
         return self.shutdown_future
 
     def is_open(self):
+        """
+        Returns:
+            bool: True if this connection is open and usable, False otherwise.
+            Check :attr:`shutdown_future` to know when the connection is completely
+            finished shutting down.
+        """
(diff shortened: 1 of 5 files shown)


source/event_stream_rpc_client_connection.c
```diff
@@ -0,0 +1,292 @@
...
+
+    aws_event_stream_rpc_client_connection_close(connection->native, AWS_OP_SUCCESS);
+    Py_RETURN_NONE;
+}
+
+PyObject *aws_py_event_stream_rpc_client_connection_is_open(PyObject *self, PyObject *args) {
+    (void)self;
+    PyObject *capsule;
+    if (!PyArg_ParseTuple(args, "O", &capsule)) {
+        return NULL;
```
DavidOgunsAWS, line 280: Should be returning false?

source/event_stream_rpc_client_connection.c
```diff
@@ -0,0 +1,292 @@
...
+    }
+
+    struct connection_binding *connection = PyCapsule_GetPointer(capsule, s_capsule_name);
+    if (!connection) {
+        return NULL;
```
DavidOgunsAWS, line 285: Return false?

source/event_stream.h
```diff
@@ -0,0 +1,13 @@
...
+#define AWS_CRT_PYTHON_EVENT_STREAM_H
+/**
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0.
+ */
+#include "module.h"
+
+PyObject *aws_py_event_stream_rpc_client_connection_connect(PyObject *self, PyObject *args);
+PyObject *aws_py_event_stream_rpc_client_connection_close(PyObject *self, PyObject *args);
+PyObject *aws_py_event_stream_rpc_client_connection_is_open(PyObject *self, PyObject *args);
```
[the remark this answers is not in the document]
graebm: The headers in this project are really minimal. I'd rather just lump all the event-stream public API together in one .h file than have a ton of .h files with like 3 function declarations each.

Building on this approach, in the Ruby CRT there's just one .h file.

source/event_stream_rpc_client_connection.c
```diff
@@ -0,0 +1,292 @@
...
+    bool shutdown_complete;
+    bool capsule_destroyed;
+
+    PyObject *on_setup;
+    PyObject *on_shutdown;
+    PyObject *on_protocol_message;
+};
+
+PyObject *aws_py_event_stream_rpc_client_connection_connect(PyObject *self, PyObject *args) {
+    (void)self;
```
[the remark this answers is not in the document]
graebm: whoa. I didn't know that

fixing...

source/event_stream_rpc_client_connection.c
graebm: For C functions bound to python, error handling works like this:
If there's an error, set the exception and return `NULL`.
In this case, PyArg_ParseTuple() has already set an exception about failure to parse the arguments passed into this fucntion.

source/event_stream_rpc_client_connection.c
graebm: nope. NULL because there's been an exception (it was set by PyCapsule_GetPointer())

source/event_stream_rpc_client_connection.c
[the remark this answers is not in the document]
graebm: wait. No there's not :|
There's an ancient PR that never got merged because no one had the will to got an update all our codebases to use it

awscrt/eventstream.py
```diff
@@ -0,0 +1,197 @@
...
+
+    @staticmethod
+    def _on_protocol_message(bound_weak_header, headers, payload, message_type, flags):
+        handler = bound_weak_handler()
+        if handler:
+            handler.on_protocol_message(
+                headers=headers,
+                payload=payload,
+                message_type=message_type,
+                flags=flags)
```
DavidOgunsAWS, line 174: Headers, payload, message type, and flag types to be understood later? All are probably built-in types except Header which needs to be bound for creation and correct interpretation in Python.

DavidOgunsAWS (approved):

awscrt/eventstream.py
graebm: yeah, this will be in next PR

PR #203, opened 2020-11-05
event-stream headers and protocol messages

graebm:
Define headers, enums, flags, etc necessary for sending and receiving protocol messages.

2 choices made for simplicity over efficiency:
1) Rather the pass complex `EventStreamHeader` class back and forth between python and C, I always pass a simple `(name, value, type)` tuple. Transforming back and forth between them is less efficient than dealing directly with the `EventStreamHeader` class in C, but it makes the C code much simpler and less likely to leak references. We can always come back and make this more efficient in the future without affecting the public API.
2) Deep-copying contents of python headers to C headers. We could get away with non-owning references to the header names and values, but the code would be more complex. Again, we can change this under the hood without affecting the public API.

### DIFF
--- a/awscrt/eventstream.py
+++ b/awscrt/eventstream.py
@@ -12,7 +12,7 @@ import awscrt.exceptions
 from awscrt.io import ClientBootstrap, SocketOptions, TlsConnectionOptions
 from collections.abc import ByteString
 from concurrent.futures import Future
-from enum import IntEnum, IntFlag
+from enum import IntEnum
 from functools import partial
 from typing import Any, Optional, Sequence
 from uuid import UUID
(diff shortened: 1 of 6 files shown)


awscrt/eventstream.py
```diff
@@ -195,3 +447,27 @@ def is_open(self):
...
         """
         return _awscrt.event_stream_rpc_client_connection_is_open(self._binding)
+
+    def send_protocol_message(
+            self,
+            *,
+            message_type: EventStreamRpcMessageType,
+            headers: Optional[Sequence[EventStreamHeader]] = [],
+            payload: Optional[ByteString] = b'',
+            flags: int = EventStreamRpcMessageFlag.NONE) -> Future:
```
DavidOgunsAWS, line 457: For consistency with Java style, the order for sending or receiving messages is always headers, payload, messageType, messageFlags.

graebm: tldr; doesn't matter

So python has the best argument parsing options in the world:
Everything after the `(self, *, ...)` is a "named keyword" argument, which is a new feature in python 3. It means the caller cannot pass these arguments by position alone, they MUST name each argument:
```
NOPE: connection.send_protocol_message(my_type, my_headers, my_payload, my_flags)
YEP:    connection.send_protocol_message(
                payload=my_payload,
                headers=my_headers,
                message_type=my_type,
                flags=my_flags)
```
The advantage of doing it this way is: you can make ALL kinds of API changes in a backwards compatible way. You can make an argument optional (which used to be required) by simply adding a default value. If the arguments were positional, that kind of change would be much more difficult, since python doesn't have function overloading.

Anyway, what I'm saying is: the order here doesn't matter. I made message_type the 1st argument because I made every other argument optional. But users can pass the arguments in any order they want, because they're naming the arguments, so the order doesn't matter.

DavidOgunsAWS (approved):

PR #204, opened 2020-11-10
Eventstream stream/continuation

graebm:
Also, some changes to the connection:
- Connection handler is now a pure abstract base class, user must store reference to connection themselves. It was weird to use the abstract-base-class stuff, but not be purely abstract
- set connection._binding from on_setup callback (instead of `connection._binding = connect()`) to avoid race condition where on_setup fires before thread calling connect() can store the binding.

### DIFF
--- a/awscrt/eventstream.py
+++ b/awscrt/eventstream.py
@@ -472,7 +472,6 @@ class EventStreamRpcClientConnection(NativeResource):
             else:
                 bound_future.set_result(None)
 
-
     @staticmethod
     def _on_connection_shutdown(bound_future, bound_weak_handler, error_code):
         reason = awscrt.exceptions.from_code(error_code) if error_code else None
--- a/awscrt/eventstream.py
+++ b/awscrt/eventstream.py
@@ -441,7 +441,7 @@ class EventStreamRpcClientConnection(NativeResource):
         # or references to futures within the connection rather than the connection itself.
         handler_weakref = weakref.ref(handler)
 
-        connection._binding = _awscrt.event_stream_rpc_client_connection_connect(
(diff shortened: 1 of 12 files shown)


awscrt/eventstream.py
```diff
@@ -445,6 +570,16 @@ def _on_connection_shutdown(bound_future, bound_weak_handler, error_code):
...
                 bound_future.set_result(None)
 
+    @staticmethod
+    def _on_continuation_closed(bound_future, bound_weak_handler):
+        try:
+            handler = bound_weak_handler()
+            if handler:
+                handler.on_continuation_closed()
+        finally:
+            # user callback had unhandled exception, use finally to ensure future gets set
```
JonathanHenson, line 580: is this comment correct?

source/event_stream_rpc_client_connection.c
```diff
@@ -137,23 +152,29 @@ static int s_on_connection_setup(
...
 
-    PyObject *result = PyObject_CallFunction(connection->on_setup, "(i)", error_code);
+    PyObject *result = PyObject_CallFunction(connection->on_setup, "(Oi)", connection->capsule_tmp, error_code);
     if (result) {
         Py_DECREF(result);
     } else {
         /* Callback might fail during application shutdown */
         PyErr_WriteUnraisable(PyErr_Occurred());
+
+        /* TODO: close connection if unhandled exception occurs? */
```
JonathanHenson, line 165: yes.

JonathanHenson (approved):

awscrt/eventstream.py
graebm: rewrote to be less confusing

PR #206, opened 2020-11-17
eventstream fixes

graebm:
- fix `setup.py` so that eventstream modules get installed
- fix crash when `eventstream.rpc.ClientConnection` was GC'd during `on_connection_shutdown` callback
- change default args to always be `=None`, so that it's easy to pass along defaults from wrapper functions
- centralize logic that sanitizes message args going in each direction between C <-> Python

### DIFF
--- a/awscrt/eventstream/rpc.py
+++ b/awscrt/eventstream/rpc.py
@@ -185,7 +185,7 @@ class ClientConnectionHandler(ABC):
         pass
 
 
-def _msg_args_for_bindings(headers, payload, message_type, flags):
+def _to_binding_msg_args(headers, payload, message_type, flags):
     """
     Transform args that a python send-msg function would take,
     into args that a native send-msg function would take.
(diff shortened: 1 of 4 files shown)


awscrt/eventstream/rpc.py
```diff
@@ -185,6 +185,35 @@ def on_protocol_message(
         pass
 
 
+def _msg_args_for_bindings(headers, payload, message_type, flags):
```
justinboswell, line 188: can this be `_msg_args_to_bindings`? `for`/`from` is going to catch someone

source/event_stream_rpc_client_continuation.c
```diff
@@ -162,20 +162,12 @@ static void s_on_continuation_message(
...
-        payload_ptr = "";
-    }
-
     PyObject *result = PyObject_CallFunction(
         continuation->on_message,
         "(Oy#iI)",
         /* NOTE: if headers_create() returns NULL, then PyObject_CallFunction() fails too, which is convenient */
         aws_py_event_stream_python_headers_create(message_args->headers, message_args->headers_count),
-        payload_ptr,
+        message_args->payload->buffer,
```
justinboswell, line 170: Good call.

justinboswell (approved):
Fix-n-ship!

PR #209, opened 2020-12-07
Update to new default resolver creation function

bretambrose:
* Updates host resolver creation to new interface

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #211, opened 2020-12-08
latest submodules to fix deadlock

graebm:
fixes possible deadlock in event-stream on_continuation_closed callbacks

also: Update to new default resolver creation function


graebm:
Nevermind. This change is already happening here: https://github.com/awslabs/aws-crt-python/pull/209

PR #212, opened 2020-12-09
S3

TingDaoK:
*Issue #, if available:*

*Description of changes:*


source/s3_client.c
```diff
@@ -0,0 +1,168 @@
...
+    PyObject *result = PyObject_CallFunction(client->on_shutdown, NULL);
+    if (result) {
+        Py_DECREF(result);
+    } else {
+        /* Callback might fail during application shutdown */
+        PyErr_WriteUnraisable(PyErr_Occurred());
+    }
+    Py_CLEAR(client->on_shutdown);
+
+    if (destroy_after_shutdown) {
```
graebm, line 65: I think you need to move the
```
client->shutdown_called = true;
bool destroy_after_shutdown = client->release_called;
```
lines down here, after you've called `Py_CLEAR(client->on_shutdown)`

Imagine this scenario:
- python object is still alive
- s_s3_client_shutdown() runs
- sets shutdown_called = true
- invokes on_shutdown
  - from on_shutdown callback, user drops reference to python object
  - now capsule destructor runs, s_s3_client_release runs, sees that shutdown_called==true and release_called==true, so destroys memory
- on_shutdown callback returns, you run `Py_CLEAR(client->on_shutdown);` but crash because client memory has been deleted

I fixed this exact same kind of bug a month ago so it's fresh in my mind

source/s3_client.c
```diff
@@ -0,0 +1,168 @@
...
+    struct aws_s3_client *native;
+
+    bool release_called;
+    bool shutdown_called;
+
+    /* Shutdown callback, reference cleared after setting result */
+    PyObject *on_shutdown;
+};
+
+static void s_s3_client_release(struct s3_client_binding *client) {
```
graebm, line 22: trivial: this fn is only ever called from the capsule_destructor. why not just combine the functions?

source/s3_client.c
```diff
@@ -0,0 +1,168 @@
...
+    s3_clinet->on_shutdown = on_shutdown_py;
+    Py_INCREF(s3_clinet->on_shutdown);
+
+    return capsule;
+
+capsule_new_failed:
+    aws_s3_client_release(s3_clinet->native);
+client_init_failed:
+    aws_mem_release(allocator, s3_clinet);
+    return NULL;
```
graebm, line 167: Ugh. setup & shutdown error-handling code is always so fragile because you need to account for setup failing halfway through. For these reasons, I've come to prefer creating the capsule as soon as possible, rather than creating it as the last step. That way, if anything goes wrong I can just decref the capsule and let its destructor clean things up, even if setup only partially succeeded. Personally, I've found it simpler to write shutdown code that can handle a partially set up thing, instead of setup code that can shut down a partially setup thing, in addition to regular shutdown code.

So the order of fallible setup operations would be like:
```
s3_client = aws_mem_calloc()
capsule = PyCapsule_New(...)
if (!capsule)
    aws_mem_destroy(s3_client)
    return NULL

s3_client->on_shutdown = on_shutdown_py
Py_INCREF(on_shutdown)

s3_client->native = aws_s3_client_new()
if (!native)
    PyErr_SetAwsLastError()
    goto error;
return capsule
error:
Py_DECREF(capsule);
return NULL
```

but cleanup needs to check for the case that the native S3 client never actually existed
```
--- s_capsule_destructor() ---
aws_s3_client_release(client->native) // yeah, may be NULL, but release() can handle it
client->release_called=true;
s_destroy_if_ready(client)

--- s_on_shutdown() ---
fire callbacks
Py_CLEAR(client->on_shutdown)
client->shutdown_called = true
s_destroy_if_ready(client)

--- s_destroy_if_ready() ---
if (client->native && (!client->shutdown_called || !client->release_called))
    // native client successfully created, but not ready to clean up yet
    return

Py_XDECREF(client->on_shutdown) // just in case native never existed and shutdown never happened
aws_mem_release(client)
```

awscrt/s3.py
```diff
@@ -0,0 +1,204 @@
...
+
+    def _on_body(self, chunk, offset):
+        if self._on_body_cb:
+            self._on_body_cb(chunk=chunk, offset=offset, kwargs=self._kwargs)
+
+    def _on_finish(self, error_code):
+        if error_code:
+            self.finished_future.set_exception(awscrt.exceptions.from_code(error_code))
+        else:
+            self.finished_future.set_result("finish")
```
graebm, line 198: trivial: In HTTP, we set a similar future's result to be the `status_code`
could do the same here

awscrt/s3.py
```diff
@@ -0,0 +1,204 @@
...
+            on_headers=on_headers,
+            on_body=on_body,
+            on_done=on_done,
+            **kwargs)
+
+
+class S3Request(NativeResource):
+    """S3 request
+
+    """
```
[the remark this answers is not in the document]
graebm: also document the `shutdown_event` attribute

test/test_s3.py
```diff
@@ -0,0 +1,134 @@
...
+        self.assertTrue(shutdown_event.wait(self.timeout))
+
+
+class S3RequestTest(NativeResourceTest):
+    get_test_object_path = "/get_object_test_10MB.txt"
+    put_test_object_path = "/put_object_test_py_10MB.txt"
+    region = "us-west-2"
+    bucket_name = "aws-crt-canary-bucket"
+    timeout = 100  # seconds
+    num_threads = 0
```
graebm, line 62: trivial: class constants are usually all caps like `REGION = "us-west-2"`

test/test_s3.py
```diff
@@ -0,0 +1,134 @@
...
+
+    response_headers = None
+    response_status_code = None
+    received_body_len = 0
+
+    put_body_stream = None
```
graebm, line 68: instance attributes are usually set in an init fn like:
```
def __init__(self):
    super().__init__()
    self.response_headers = None
    ...
```

test/test_s3.py
```diff
@@ -0,0 +1,134 @@
...
+        return request
+
+    def _on_request_headers(self, status_code, headers, **kargs):
+        self.response_status_code = status_code
+        self.assertIsNotNone(headers, "headers are none")
+        self.response_headers = headers
+
+    def _on_request_body(self, chunk, offset, **kargs):
+        self.assertIsNotNone(chunk, "the body chunk is none")
+        self.received_body_len = self.received_body_len + len(chunk)
```
graebm, line 94: having an assert here isn't going to do anything, these assert functions just raise exceptions
But this is on a callback from C on another thread
and as I mentioned earlier, exceptions on these callbacks don't do anything to stop the operation. PyErr_WriteUnraisable() will write something to stderr, but everything is going to keep charging along

the pattern I've been using for detecting errors in void callbacks is to do like:
```
    if chunk is None:
        self.callback_error = "the body chunk is none"
```

then at the end of the test on the main thread, after all the callbacks have fired, I check if any of them reported an error, like: `self.assertIsNone(self.callback_error)`

test/test_s3.py
```diff
@@ -0,0 +1,134 @@
...
+            request=request,
+            type=type,
+            on_headers=self._on_request_headers,
+            on_body=self._on_request_body)
+        finished_future = s3_request.finished_future
+        finished_future.result(self.timeout)
+        self._validate_successful_get_response(type is AwsS3RequestType.PUT_OBJECT)
+        shutdown_event = s3_request.shutdown_event
+        del s3_request
+        self.assertTrue(shutdown_event.wait(self.timeout))
```
graebm, line 121: nice job btw
I saw how carefully you're setting the `shutdown_event` without affecting the refcount of anything else

source/s3_client.c
```diff
@@ -0,0 +1,168 @@
...
+    }
+
+    struct aws_client_bootstrap *bootstrap = aws_py_get_client_bootstrap(bootstrap_py);
+    if (!bootstrap) {
+        return NULL;
+    }
+
+    struct aws_credentials_provider *credential_provider = NULL;
+    if (credential_provider_py != Py_None) {
+        credential_provider = aws_py_get_credentials_provider(credential_provider_py);
```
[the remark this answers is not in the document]
TingDaoK: Oh, so the python object is not none, `aws_py_get_credentials_provider` should have a non-null returned value?

awscrt/s3.py
TingDaoK: feels like we can just set no result since the invalid response status code will result in a failure from native client.

awscrt/s3.py
```diff
@@ -0,0 +1,204 @@
...
+
+    def __init__(
+            self,
+            *,
+            bootstrap,
+            region,
+            credential_provider=None,
+            tls_connection_options=None,
+            part_size=0,
+            throughput_target_gbps=0):
```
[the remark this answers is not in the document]
TingDaoK: seems like we are doing the same thing for mqtt. So, if we set the default to None, we need to treat it as an object in the C land?

source/s3_meta_request.c
```diff
@@ -0,0 +1,277 @@
...
+
+    bool destroy_after_shutdown = request_binding->release_called;
+
+    /* Invoke on_shutdown, then clear our reference to it */
+    PyObject *result = PyObject_CallFunction(request_binding->on_shutdown, NULL);
+    if (result) {
+        Py_DECREF(result);
+    } else {
+        /* Callback might fail during application shutdown */
+        PyErr_WriteUnraisable(PyErr_Occurred());
```
TingDaoK, line 185: self_py has been cleared here

awscrt/auth.py
```diff
@@ -255,6 +255,32 @@ def new_chain(cls, providers):
         binding = _awscrt.credentials_provider_new_chain(providers)
         return cls(binding)
 
+    @classmethod
+    def new_python(cls, py_provider):
```
JonathanHenson, line 259: I don't love this name. Maybe "new_custom" or something? I dunno. @graebm is the naming hero around her.

awscrt/auth.py
```diff
@@ -255,6 +255,32 @@ def new_chain(cls, providers):
...
+        Creates a provider that sources credentials from a python class.
+
+        The python provider needs to have a method :get_credential(), which returns
+        the credential as a dictionary like (Expiration and SessionToken are optional):
+        {
+            "AccessKeyId": "accesskey",
+            "SecretAccessKey": "secretAccessKey",
+            "SessionToken": "....",
+            "Expiration": 1247169778 // timepoint, in seconds since epoch
+        }
```
JonathanHenson, line 270: how will this work with something like ECDSA keys? @bretambrose how you feel about this one?

awscrt/s3.py
```diff
@@ -0,0 +1,282 @@
...
+    PUT_OBJECT = 2
+    """
+    Put Object S3 request
+    """
+
+
+class S3RequestTlsMode(IntEnum):
+    """Tls mode for S3 request"""
+
+    AWS_MR_TLS_ENABLED = 0
```
JonathanHenson, line 41: what's MR?

s3_benchmark.py
```diff
@@ -0,0 +1,179 @@
...
+        self.last_sample_time = time.time()
+
+    def record_read(self, size):
+        with self._lock:
+            self._bytes_read += size
+            if self.sec_first_byte == 0:
+                self.sec_first_byte = time.time() - self.star_time
+            time_now = time.time()
+            if time_now - self.last_sample_time > 1:
+                bytes_this_second = (self._bytes_read - self._bytes_sampled) / (time_now - self.last_sample_time)
```
JonathanHenson, line 75: We probably should batch these before even calling into python in the first place.

s3_benchmark.py
```diff
@@ -0,0 +1,179 @@
...
+object_real_name = "/0_10GB"
+suffix = ".txt"
+repeat_times = 1
+bunch_size = 160
+
+writing_disk = True
+request_type = "download"
+
+# Initialization
+event_loop_group = EventLoopGroup(18)
```
JonathanHenson, line 103: can we bind out the new numa aware event-loop-groups here and give this a go on node 0?

s3_benchmark.py
```diff
@@ -0,0 +1,179 @@
...
+# data_len = file_stats.st_size
+
+# data_stream = CrtLazyReadStream(file_name, "r+b", t_statistic, data_len)
+# upload_headers = HttpHeaders([("host", bucket_name + ".s3." + region + ".amazonaws.com"),
+#                               ("Content-Type", "text/plain"), ("Content-Length", str(data_len))])
+# upload_request = HttpRequest("PUT", "/put_object_test_py_10MB.txt", upload_headers, data_stream)
+
+
+def on_body(offset, chunk, **kwargs):
+    t_statistic.record_read(len(chunk))
```
JonathanHenson, line 128: isn't chunk NULL when using the FILE* api?

source/s3_meta_request.c
```diff
@@ -0,0 +1,335 @@
...
+    struct aws_s3_meta_request *meta_request,
+    const struct aws_http_headers *headers,
+    int response_status,
+    void *user_data) {
+    (void)meta_request;
+    struct s3_meta_request_binding *request_binding = user_data;
+
+    /*************** GIL ACQUIRE ***************/
+    PyGILState_STATE state;
+    if (aws_py_gilstate_ensure(&state)) {
```
JonathanHenson, line 84: Do we have to have the GIL at this point? could we delay it until we're ready to perform the callback?

source/s3_meta_request.c
```diff
@@ -0,0 +1,335 @@
...
+        return; /* Python has shut down. Nothing matters anymore, but don't crash */
+    }
+
+    PyObject *result = PyObject_CallMethod(
```
JonathanHenson, line 129: this will be slow as hell under heavy concurrency load.

s3_benchmark.py
TingDaoK: The on_body callback will not be invoked when the FILE pointer is set.

awscrt/s3.py
TingDaoK: Copy/pasted from https://github.com/awslabs/aws-c-s3/blob/main/include/aws/s3/s3_client.h#L54-L57.
Should stand for meta_request. Hmm, since we are not taking meta request to python land, we probably should get out of it🤔

source/s3_client.h
```diff
@@ -0,0 +1,17 @@
+#ifndef AWS_CRT_PYTHON_S3_CLIENT_H
```
graebm, line 1: trivial: maybe just name this file s3.h instead of s3_client.h
It's simpler to just have 1 header per submodule

source/s3_client.h
```diff
@@ -0,0 +1,17 @@
...
+#define AWS_CRT_PYTHON_S3_CLIENT_H
+/**
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0.
+ */
+
+#include "module.h"
+
+#include <aws/s3/s3_client.h>
```
graebm, line 10: trivial: don't need to include anything but module.h in this header file

source/s3_client.c
```diff
@@ -0,0 +1,171 @@
...
+};
+
+static void s_destroy_if_ready(struct s3_client_binding *client) {
+    if (client->native && (!client->shutdown_called || !client->release_called)) {
+        /* native client successfully created, but not ready to clean up yet */
+        return;
+    }
+    /* in case native never existed and shutdown never happened */
+    Py_XDECREF(client->on_shutdown);
+    aws_mem_release(aws_py_get_allocator(), client);
```
graebm, line 29: trivial: just delete the "in case native never existed" comment. this code path is also taken if native DID exist and normal shutdown is complete. The comment above this, about "not ready to clean up yet" conveys this all pretty well

source/s3_client.c
```diff
@@ -0,0 +1,171 @@
...
+    AWS_FATAL_ASSERT(!client->release_called);
+
+    aws_s3_client_release(client->native);
+    client->release_called = true;
+
+    s_destroy_if_ready(client);
+}
+
+/* Callback from C land, invoked when the underlying shutdown process finished */
+static void s_s3_client_shutdown(void *user_data) {
```
graebm, line 44: 🤣
"C land"
I love it

graebm (commented):
all trivial feedback. Going to wait until this is merged with the other PRs before reviewing further

graebm (commented):
still reading, but here's some more feedback

source/s3_meta_request.c
```diff
@@ -0,0 +1,673 @@
...
+        aws_input_stream_destroy(meta_request->input_body);
+    }
+
+    if (meta_request->copied_message) {
+        aws_http_message_release(meta_request->copied_message);
+    }
+    /* in case native never existed and shutdown never happened */
+    Py_XDECREF(meta_request->on_shutdown);
+    Py_XDECREF(meta_request->py_core);
+    aws_mem_release(aws_py_get_allocator(), meta_request);
```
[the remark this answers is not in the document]
TingDaoK: `s_s3_meta_request_capsule_destructor` cleaned those up. So, if anything fails, the capsule destructor should still be invoked, right?

source/s3_meta_request.c
```diff
@@ -0,0 +1,673 @@
...
+    if (result) {
+        Py_DECREF(result);
+    } else {
+        /* Callback might fail during application shutdown */
+        PyErr_WriteUnraisable(request_binding->self_py);
+    }
+done:
+    Py_XDECREF(header_list);
+    Py_CLEAR(request_binding->self_py);
+    Py_CLEAR(request_binding->http_message_py);
```
graebm: you're might remove the http_message_py variable from existence anyway...

but if you keep it, this isn't the right place to clear it
it still needs to exist until the shutdown callback fires, right?
because if we cancel other threads might still be reading from it?

source/s3_meta_request.c
```diff
@@ -0,0 +1,673 @@
...
+
+    /*************** GIL ACQUIRE ***************/
+    PyGILState_STATE state;
+    if (aws_py_gilstate_ensure(&state)) {
+        return; /* Python has shut down. Nothing matters anymore, but don't crash */
+    }
+
+    request_binding->shutdown_called = true;
+
+    bool destroy_after_shutdown = request_binding->release_called;
```
graebm: is it possible for `destroy_after_shutdown` to ever be false?

if it's not possible, let's not even have this variable

source/s3_meta_request.c
```diff
@@ -0,0 +1,673 @@
...
+    } else {
+        /* Callback might fail during application shutdown */
+        PyErr_WriteUnraisable(PyErr_Occurred());
+    }
+    Py_CLEAR(request_binding->on_shutdown);
+    Py_CLEAR(request_binding->py_core);
+
+    if (destroy_after_shutdown) {
+        aws_mem_release(aws_py_get_allocator(), request_binding);
+    }
```
graebm: shouldn't we just be calling `s_destroy_if_ready()`?
having cleanup code in many different places is ... error prone as time goes on and more people touch this code

source/s3_meta_request.c
```diff
@@ -0,0 +1,673 @@
...
+    struct s3_meta_request_binding *request_binding = impl->binding;
+    if (s_record_progress(request_binding, (uint64_t)actually_read, &report_progress)) {
+        return AWS_OP_ERR;
+    }
+
+    if (report_progress) {
+        /*************** GIL ACQUIRE ***************/
+        PyGILState_STATE state;
+        if (aws_py_gilstate_ensure(&state)) {
+            return AWS_OP_SUCCESS; /* Python has shut down. Nothing matters anymore, but don't crash */
```
graebm, line 366: return AWS_OP_ERR;

source/s3_meta_request.c
```diff
@@ -0,0 +1,673 @@
...
+        }
+        PyObject *result =
+            PyObject_CallMethod(request_binding->self_py, "_on_progress", "(K)", request_binding->size_transferred);
+        if (!result) {
+            return aws_py_raise_error();
```
graebm, line 371: you can't `return`! We're holding the GIL!

source/s3_meta_request.c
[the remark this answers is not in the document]
graebm: yeah, I saw that later. cool

source/s3_meta_request.c
```diff
@@ -0,0 +1,673 @@
...
+    if (error) {
+        return aws_raise_error(AWS_ERROR_CRT_CALLBACK_EXCEPTION);
+    } else {
+        return AWS_OP_SUCCESS;
+    }
+}
+
+/* If the request has not finished, it will keep the request alive, until the finish callback invoked. So, we don't
+ * need to clean anything from this call */
+static void s_s3_request_on_finish(
```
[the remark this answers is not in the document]
TingDaoK: eh... I am confused by myself🤯. Probably related to some crazy lifetime management before. Will remove!

source/s3_meta_request.c
```diff
@@ -0,0 +1,673 @@
...
+        error_body = *(meta_request_result->error_response_body);
+    }
+    result = PyObject_CallMethod(
+        request_binding->self_py,
+        "_on_finish",
+        "(iOy#)",
+        meta_request_result->error_code,
+        header_list ? header_list : Py_None,
+        (const char *)(error_body.buffer),
+        (Py_ssize_t)error_body.len);
```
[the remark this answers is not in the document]
TingDaoK: I think it's probably better to have either `None` or `bytes`? Since the headers can be `None` as well. Will document it!

source/s3_meta_request.c
TingDaoK: Yep, removed!

source/s3_meta_request.c
```diff
@@ -0,0 +1,623 @@
...
+    if (error) {
+        return aws_raise_error(AWS_ERROR_CRT_CALLBACK_EXCEPTION);
+    } else {
+        return AWS_OP_SUCCESS;
+    }
+}
+
+static int s_record_progress(struct s3_meta_request_binding *request_binding, uint64_t length, bool *report_progress) {
+    if (aws_add_u64_checked(request_binding->size_transferred, length, &request_binding->size_transferred)) {
+        /* Wow */
```
graebm, line 132: 🤣

graebm (approved):
beautiful!

PR #214, opened 2020-12-14
profile credential provider binding

TingDaoK:
*Issue #, if available:*

*Description of changes:*


justinboswell (approved):

awscrt/auth.py
```diff
@@ -140,6 +140,28 @@ def new_static(cls, access_key_id, secret_access_key, session_token=None):
...
 
+    @classmethod
+    def new_profile(
+            cls,
+            client_bootstrap,
+            profile_name='default',
+            config_file_name='~/.aws/config',
+            credentials_file_name='~/.aws/credentials'):
+        """
+        Create a provider only search for profile.
```
graebm, line 151: Use the docs from aws-c-auth's .h file:
```
        Creates a provider that sources credentials from key-value profiles
        loaded from the aws credentials file ("~/.aws/credentials" by default)
        and the aws config file ("~/.aws/config" by default)
```

awscrt/auth.py
```diff
@@ -140,6 +140,28 @@ def new_static(cls, access_key_id, secret_access_key, session_token=None):
         binding = _awscrt.credentials_provider_new_static(access_key_id, secret_access_key, session_token)
         return cls(binding)
 
+    @classmethod
+    def new_profile(
+            cls,
+            client_bootstrap,
+            profile_name='default',
+            config_file_name='~/.aws/config',
+            credentials_file_name='~/.aws/credentials'):
```
graebm, line 149: For the last 3 args, have them be `=None` by default

On windows "~" isn't a thing, so we can't use it as a default string here.
Down in C, you're allowed to passed zeroed-out byte-cursors to get the default values, so take advantage of that and let C deal with the platform-specific home directory stuff

source/auth_credentials.c
```diff
@@ -120,15 +120,10 @@ PyObject *aws_py_credentials_session_token(PyObject *self, PyObject *args) {
  */
 struct credentials_provider_binding {
     struct aws_credentials_provider *native;
-
-    /* Dependencies that must outlive this.
-     * Note that different types of providers have different dependencies */
-    PyObject *bootstrap;
```
graebm, line 126: nice cleanup 👍

source/auth_credentials.c
```diff
@@ -340,3 +332,59 @@ PyObject *aws_py_credentials_provider_new_static(PyObject *self, PyObject *args)
...
+    struct aws_allocator *allocator = aws_py_get_allocator();
+
+    PyObject *bootstrap_py;
+    struct aws_byte_cursor profile_name;
+    struct aws_byte_cursor config_file_name;
+    struct aws_byte_cursor credentials_file_name;
+
+    if (!PyArg_ParseTuple(
+            args,
+            "Os#s#s#",
```
graebm, line 347: you can do `z#` instead of `s#`
then if `None` comes from Python, then the ptr will be set NULL and the len will be set 0
it's quite convenient

test/test_auth.py
```diff
@@ -101,6 +101,44 @@ def test_default_provider(self):
             self.assertTrue(EXAMPLE_SECRET_ACCESS_KEY == credentials.secret_access_key)
             self.assertTrue(credentials.session_token is None)
 
+    def _prepare_profile_file(self, profile_name, credential_file_name):
+        # prepare a credential profile file for the credential provider
+        file = open(credential_file_name, "w")
+        file.write("[default]\n" +
```
graebm, line 107: Seems simpler to just commit a real actual file to `test/resources/example_profile`?

If you need to write the file from code, use something from [tempfile](https://docs.python.org/3/library/tempfile.html) to create a temporary file in the OS's appropriate tmp directory

graebm (approved):

PR #215, opened 2020-12-18
python credential provider

TingDaoK:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/awscrt/auth.py
+++ b/awscrt/auth.py
@@ -76,6 +76,7 @@ class AwsCredentialsProviderBase(NativeResource):
     # AwsCredentialsProvider.new_delegate() instead.
     pass
 
+
 class AwsCredentialsProvider(AwsCredentialsProviderBase):
     """
     Credentials providers source the AwsCredentials needed to sign an authenticated AWS request.
--- a/awscrt/auth.py
+++ b/awscrt/auth.py
@@ -256,7 +256,7 @@ class AwsCredentialsProvider(AwsCredentialsProviderBase):
         return cls(binding)
 
     @classmethod
-    def new_py_provider(cls, py_provider):
+    def new_python(cls, py_provider):
         """
         Creates a provider that sources credentials from a python class.
 
@@ -266,17 +266,19 @@ class AwsCredentialsProvider(AwsCredentialsProviderBase):
             "AccessKeyId": "accesskey",
             "SecretAccessKey": "secretAccessKey",
             "SessionToken": "....",
-            "Expiration": 1247169778
+            "Expiration": 1247169778 // timepoint, in seconds since epoch
         }
 
         Args:
-            providers (List[AwsCredentialsProvider]): List of credentials providers.
+            provider: python Object.
 
         Returns:
             AwsCredentialsProvider:
         """
 
-        binding = _awscrt.credentials_provider_new_py_provider(py_provider)
+        assert hasattr(py_provider, 'get_credential')
+        assert callable(py_provider.get_credential)
+        binding = _awscrt.credentials_provider_new_python(py_provider)
         return cls(binding)
 
     def get_credentials(self):
--- a/awscrt/auth.py
+++ b/awscrt/auth.py
@@ -261,9 +261,8 @@ class AwsCredentialsProvider(AwsCredentialsProviderBase):
         Creates a provider that sources credentials from a python class.
 
         The python provider needs to have a method :get_credential(), which returns
-        the credential as a dictionary like:
+        the credential as a dictionary like (Expiration and SessionToken are optional):
         {
-            "Version": 1,
             "AccessKeyId": "accesskey",
             "SecretAccessKey": "secretAccessKey",
             "SessionToken": "....",
--- a/awscrt/auth.py
+++ b/awscrt/auth.py
@@ -255,6 +255,31 @@ class AwsCredentialsProvider(AwsCredentialsProviderBase):
         binding = _awscrt.credentials_provider_new_chain(providers)
         return cls(binding)
 
+    @classmethod
+    def new_py_provider(cls, py_provider):
+        """
+        Creates a provider that sources credentials from a python class.
+
+        The python provider needs to have a method :get_credential(), which returns
+        the credential as a dictionary like:
+        {
+            "Version": 1,
+            "AccessKeyId": "accesskey",
+            "SecretAccessKey": "secretAccessKey",
+            "SessionToken": "....",
+            "Expiration": 1247169778
+        }
+
+        Args:
+            providers (List[AwsCredentialsProvider]): List of credentials providers.
+
+        Returns:
+            AwsCredentialsProvider:
+        """
+
+        binding = _awscrt.credentials_provider_new_py_provider(py_provider)
+        return cls(binding)
+
     def get_credentials(self):
         future = Future()
 
--- a/awscrt/auth.py
+++ b/awscrt/auth.py
@@ -67,43 +67,27 @@ class AwsCredentials(NativeResource):
 
 
 class AwsCredentialsProviderBase(NativeResource):
-    """
-    Base class for providers that source the AwsCredentials needed to sign an authenticated AWS request.
-
-    NOTE: Custom subclasses of AwsCredentialsProviderBase are not yet supported.
-    """
-    __slots__ = ()
-
-    def __init__(self, binding=None):
-        super().__init__()
-
-        if binding is None:
-            # TODO: create binding type that lets native code call into python subclass
-            raise NotImplementedError("Custom subclasses of AwsCredentialsProviderBase are not yet supported")
-
-        self._binding = binding
-
-    def get_credentials(self):
-        """
-        Asynchronously fetch AwsCredentials.
-
-        Returns:
-            concurrent.futures.Future: A Future which will contain
-            :class:`AwsCredentials` (or an exception) when the operation completes.
-            The operation may complete on a different thread.
-        """
-        raise NotImplementedError()
-
+    # Pointless base class, kept for backwards compatibility.
+    # AwsCredentialsProvider is (and always will be) the only subclass.
+    #
+    # Originally created with the thought that, when we supported
+    # custom python providers, they would inherit from this class.
+    # We ended up supporting custom python providers via
+    # AwsCredentialsProvider.new_delegate() instead.
+    pass
 
 class AwsCredentialsProvider(AwsCredentialsProviderBase):
     """
     Credentials providers source the AwsCredentials needed to sign an authenticated AWS request.
 
-    Base class: AwsCredentialsProviderBase
-
     This class provides `new()` functions for several built-in provider types.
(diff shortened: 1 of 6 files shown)


awscrt/auth.py
```diff
@@ -255,6 +255,30 @@ def new_chain(cls, providers):
...
+        """
+        Creates a provider that sources credentials from a python class.
+
+        The python provider needs to have a method :get_credential(), which returns
+        the credential as a dictionary like (Expiration and SessionToken are optional):
+        {
+            "AccessKeyId": "accesskey",
+            "SecretAccessKey": "secretAccessKey",
+            "SessionToken": "....",
+            "Expiration": 1247169778
```
bretambrose, line 269: Add a note that explains expiration's time properties (units, base line, etc...)

awscrt/auth.py
```diff
@@ -255,6 +255,30 @@ def new_chain(cls, providers):
         binding = _awscrt.credentials_provider_new_chain(providers)
         return cls(binding)
 
+    @classmethod
+    def new_py_provider(cls, py_provider):
```
bretambrose, line 259: consider adding an optional (default = true) argument to decorate with a caching provider

awscrt/auth.py
```diff
@@ -255,6 +255,30 @@ def new_chain(cls, providers):
...
+        the credential as a dictionary like (Expiration and SessionToken are optional):
+        {
+            "AccessKeyId": "accesskey",
+            "SecretAccessKey": "secretAccessKey",
+            "SessionToken": "....",
+            "Expiration": 1247169778
+        }
+
+        Args:
+            providers (List[AwsCredentialsProvider]): List of credentials providers.
```
bretambrose, line 273: this seems inaccurate

awscrt/auth.py
bretambrose, line 259: maybe new_managed or new_python, to follow the pattern of the previous ones

source/auth_credentials.c
```diff
@@ -544,3 +548,124 @@ PyObject *aws_py_credentials_provider_new_chain(PyObject *self, PyObject *args)
...
+    aws_credentials_release(credentials);
+    aws_string_destroy(session_token);
+    aws_string_destroy(secret_access_key);
+    aws_string_destroy(access_key_id);
+
+    return AWS_OP_SUCCESS;
+}
+
+static void s_credentials_provider_py_provider_destroy(struct aws_credentials_provider *provider) {
+    if (provider && provider->shutdown_options.shutdown_callback) {
```
bretambrose, line 616: I would switch these statements safely (save off the callback and user data in locals).  It's possible for tests to be gated on this completing and if they do a mem check before the release completes, the mem check will fail.

source/auth_credentials.c
```diff
@@ -544,3 +548,124 @@ PyObject *aws_py_credentials_provider_new_chain(PyObject *self, PyObject *args)
...
+    uint64_t expiration = UINT64_MAX;
+    if (py_expiration) {
+        expiration = PyLong_AsUnsignedLongLong(py_expiration);
+    }
+    PyGILState_Release(state);
+    if (access_key_id != NULL && secret_access_key != NULL) {
+        credentials =
+            aws_credentials_new_from_string(allocator, access_key_id, secret_access_key, session_token, expiration);
+        if (credentials == NULL) {
+            error_code = aws_last_error();
```
bretambrose, line 599: if this is AWS_ERROR_SUCCESS, promote to INVALID_SOMETHING too

awscrt/auth.py
TingDaoK: Still not fully understand the cache provider... Can the caching provider be added as another provider in the chain provider list?

awscrt/auth.py
TingDaoK: Whoops. good catch!

awscrt/auth.py
bretambrose: Yes.

TingDaoK: So, why we need to add an option here?🤔

source/auth_credentials.c
TingDaoK: did not get what you mean by "save off the callback and user data in locals" here...

bretambrose: if you release the memory you can't use it anymore so values from it need to be pulled off first.

awscrt/auth.py
bretambrose: I guess you don't need to but then the user needs to worry about all the caching themslves.  They have no way of building a provider chain nor a caching provider, so it will all have to come from python.

awscrt/auth.py
```diff
@@ -255,6 +255,32 @@ def new_chain(cls, providers):
         binding = _awscrt.credentials_provider_new_chain(providers)
         return cls(binding)
 
+    @classmethod
+    def new_python(cls, py_provider):
```
TingDaoK, line 259: Any preference about this taking an object or a function instead? To make it simple, it may be better to be a simple function? @graebm

graebm, line 259: naming: not a fan of using the word "python" to distinguish something ... in python. I would just name this `def new_delegate(get_credentials)`

graebm: I like this taking a function, that's much more common in python-land

awscrt/auth.py
```diff
@@ -255,6 +255,32 @@ def new_chain(cls, providers):
...
+        Creates a provider that sources credentials from a python class.
+
+        The python provider needs to have a method :get_credential(), which returns
+        the credential as a dictionary like (Expiration and SessionToken are optional):
+        {
+            "AccessKeyId": "accesskey",
+            "SecretAccessKey": "secretAccessKey",
+            "SessionToken": "....",
+            "Expiration": 1247169778 // timepoint, in seconds since epoch
+        }
```
graebm, line 270: Why return a dictionary when we can have them return an actual `AwsCredentials` class?
That would eliminate the need for the long s_credentials_provider_py_provider_get_credentials() function later

We would just need to modify `AwsCredentials.__init__()` so it can accept an expiration

awscrt/auth.py
```diff
@@ -255,6 +255,32 @@ def new_chain(cls, providers):
         binding = _awscrt.credentials_provider_new_chain(providers)
         return cls(binding)
 
+    @classmethod
+    def new_python(cls, py_provider):
+        """
+        Creates a provider that sources credentials from a python class.
```
graebm, line 261: Not necessary at this time, but in the future we should support async delegate get_credentials functions.
Like, if the user return an `AwsCredentials` then we'll do what we're doing here, but if they return a `Future` then we add a done_callback to the future that passes the ultimate result down to C.

source/auth_credentials.c
```diff
@@ -120,10 +120,14 @@ PyObject *aws_py_credentials_session_token(PyObject *self, PyObject *args) {
...
 struct credentials_provider_binding {
     struct aws_credentials_provider *native;
+    PyObject *py_provider;
 };
 
 /* Finally clean up binding (after capsule destructor runs and credentials provider shutdown completes) */
 static void s_credentials_provider_binding_clean_up(struct credentials_provider_binding *binding) {
+    if (binding->py_provider) {
+        Py_XDECREF(binding->py_provider);
+    }
```
graebm, line 130: trivial: don't need the `if` here. Py_XDECREF does the if-check for you:

Py_DECREF: no if-check, always decrefs, will crash if variable is NULL
Py_XDECREF: if-check variable before decref
Py_CLEAR: if-check variable before decref, then sets variable to NULL

awscrt/auth.py
```diff
@@ -255,6 +255,32 @@ def new_chain(cls, providers):
         binding = _awscrt.credentials_provider_new_chain(providers)
```
graebm, line 255: there are a bunch of comments in the `AwsCredentialsProviderBase` class about how we don't support custom subclasses. We can remove them now. In fact, there's no need anymore for that base class anymore. For backwards compatibility, we can't remove its existence, but maybe we can remove the docstrings so it's omitted from the documentation.

I'd originally thought we'd have python programmers subclass `AwsCredentialsProviderBase` to make custom providers. But there are complex refcounting issues with that approach, the way you did it with the separate delegate object is nice and simple and graceful.

awscrt/auth.py
```diff
@@ -306,7 +319,7 @@ def _on_complete(error_code, binding):
         try:
             _awscrt.credentials_provider_get_credentials(self._binding, _on_complete)
         except Exception as e:
-            future.set_result(e)
+            future.set_exception(e)
```
TingDaoK, line 322: good catch

source/auth_credentials.c
```diff
@@ -559,3 +565,103 @@ PyObject *aws_py_credentials_provider_new_chain(PyObject *self, PyObject *args)
...
+            AWS_LS_AUTH_CREDENTIALS_PROVIDER,
+            "(id=%p) Exception in get_credentials() delegate callback",
+            (void *)binding->native);
+
+        PyErr_WriteUnraisable(binding->py_delegate);
+        goto done;
+    }
+
+    /* Expect py_result to be AwsCredentials (which wraps native aws_credentials). */
+    native_credentials = aws_py_get_credentials(py_result);
```
TingDaoK, line 596: yeah, return a credential object is making our life easier!

source/auth_credentials.c
```diff
@@ -559,3 +565,103 @@ PyObject *aws_py_credentials_provider_new_chain(PyObject *self, PyObject *args)
...
+            AWS_LS_AUTH_CREDENTIALS_PROVIDER,
+            "(id=%p) get_credentials() delegate callback failed to return AwsCredentials",
+            (void *)binding->native);
+
+        PyErr_WriteUnraisable(binding->py_delegate);
+        goto done;
+    }
+
+    /* Keep native aws_credentials alive until we pass them to callback. */
+    aws_credentials_acquire(native_credentials);
```
TingDaoK, line 608: wow, never really thought of this.🤯

TingDaoK (commented):
Oh, I cannot approve the PR I created. Ship it🚢

source/auth_credentials.c
graebm: It wasn't an issue that you needed to worry about, given the way you were doing it. It's only an issue because I changed the delegate to return a python `AwsCredentials`.
- I need to invoke the C callback after releasing the GIL
- I need to eventually DECREF the python `AwsCredentials` that got returned.
- So I'm doing this to keep the C part alive, then DECREF the python part, then release the GIL, then invoke the callback, then release() the C part

awscrt/auth.py
graebm: oye, this bug has been there since forever. I guess I never tested the failure case until now 😬

awscrt/auth.py
graebm: me and Bret chatted.
We'll add the cached_provider in a separate PR

graebm (approved):

PR #216, opened 2020-12-22
File handle from C

TingDaoK:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/awscrt/s3.py
+++ b/awscrt/s3.py
@@ -137,6 +137,7 @@ class S3Client(NativeResource):
 
         Args:
             request (HttpRequest): The overall outgoing API request for S3 operation.
+                If the request body is a file, set send_filepath for better performance.
 
             type (S3RequestType): The type of S3 request passed in, GET_OBJECT/PUT_OBJECT can be accelerated
 
@@ -144,11 +145,15 @@ class S3Client(NativeResource):
                 AwsCredentials needed to sign an authenticated AWS request, for this request only.
                 If None is provided, the credential provider in the client will be used.
 
-            recv_filepath (Optional[str]): Optional file path. If set, the C part will handle writing from a file.
-                The performance can be improved. on_body callback will not be invoked once recv_filepath is set.
+             recv_filepath (Optional[str]): Optional file path. If set, the
+                response body is written directly to a file and the
+                on_body callback is not invoked. This should give better
+                performance than writing to file from the on_body callback.
 
-            send_filepath (Optional[str]): Optional file path. If set, the C part will handle reading from a file.
-                The performance can be improved.
+            send_filepath (Optional[str]): Optional file path. If set, the
+                request body is read directly from a file and the
+                request's body_stream is ignored. This should give better
+                performance than reading a file from a stream.
 
             on_headers: Optional callback invoked as the response received, and even the API request
                 has been split into multiple parts, this callback will only be invoked once as
@@ -163,7 +168,7 @@ class S3Client(NativeResource):
                 *   `**kwargs` (dict): Forward-compatibility kwargs.
 
             on_body: Optional callback invoked 0+ times as the response body received from S3 server.
-                Once the filepath is set, this callback will never get invoked.
+                If simply writing to a file, use recv_filepath instead of on_body for better performance.
                 The function should take the following arguments and return nothing:
 
                 *   `chunk` (buffer): Response body data (not necessarily
--- a/awscrt/s3.py
+++ b/awscrt/s3.py
@@ -229,8 +229,6 @@ class S3Request(NativeResource):
             region=None):
         assert isinstance(client, S3Client)
         assert isinstance(request, HttpRequest)
-        assert file is None or os.path.isfile(file)
-        # assert file is path or file is None
         assert callable(on_headers) or on_headers is None
         assert callable(on_body) or on_body is None
         assert callable(on_done) or on_done is None
--- a/awscrt/s3.py
+++ b/awscrt/s3.py
@@ -126,7 +126,8 @@ class S3Client(NativeResource):
             request,
             type,
             credential_provider=None,
-            file=None,
+            recv_filepath=None,
+            send_filepath=None,
             on_headers=None,
             on_body=None,
             on_done=None,
@@ -143,8 +144,11 @@ class S3Client(NativeResource):
                 AwsCredentials needed to sign an authenticated AWS request, for this request only.
                 If None is provided, the credential provider in the client will be used.
 
-            file (Optional[path]): Optional file path. If set, the C part will handle writing/reading from a file.
-                The performance can be improved with it, but the memory usage will be higher
+            recv_filepath (Optional[str]): Optional file path. If set, the C part will handle writing from a file.
+                The performance can be improved. on_body callback will not be invoked once recv_filepath is set.
+
+            send_filepath (Optional[str]): Optional file path. If set, the C part will handle reading from a file.
+                The performance can be improved.
 
             on_headers: Optional callback invoked as the response received, and even the API request
                 has been split into multiple parts, this callback will only be invoked once as
@@ -159,6 +163,7 @@ class S3Client(NativeResource):
                 *   `**kwargs` (dict): Forward-compatibility kwargs.
 
             on_body: Optional callback invoked 0+ times as the response body received from S3 server.
+                Once the filepath is set, this callback will never get invoked.
                 The function should take the following arguments and return nothing:
 
                 *   `chunk` (buffer): Response body data (not necessarily
@@ -192,7 +197,8 @@ class S3Client(NativeResource):
             request=request,
             type=type,
             credential_provider=credential_provider,
-            file=file,
+            recv_filepath=recv_filepath,
+            send_filepath=send_filepath,
             on_headers=on_headers,
             on_body=on_body,
             on_done=on_done,
@@ -231,7 +237,8 @@ class S3Request(NativeResource):
             request,
             type,
             credential_provider=None,
-            file=None,
+            recv_filepath=None,
+            send_filepath=None,
             on_headers=None,
             on_body=None,
             on_done=None,
@@ -269,7 +276,8 @@ class S3Request(NativeResource):
             request,
             type,
             credential_provider,
-            file,
+            recv_filepath,
+            send_filepath,
             region,
             on_shutdown)
 
@@ -277,9 +285,9 @@ class S3Request(NativeResource):
         if self._on_headers_cb:
             self._on_headers_cb(status_code=status_code, headers=headers)
 
-    def _on_body(self, chunk, offset, size=None):
+    def _on_body(self, chunk, offset):
         if self._on_body_cb:
-            self._on_body_cb(chunk=chunk, offset=offset, size=size)
+            self._on_body_cb(chunk=chunk, offset=offset)
 
     def _on_finish(self, error_code, error_headers, error_body):
         error = None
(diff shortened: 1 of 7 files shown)


source/s3_meta_request.c
```diff
@@ -114,9 +119,27 @@ static void s_s3_request_on_body(
...
-    PyObject *result = PyObject_CallMethod(
-        request_binding->self_py, "_on_body", "(y#K)", (const char *)(body->ptr), (Py_ssize_t)body->len, range_start);
+    PyObject *result = NULL;
+    if (request_binding->file_py) {
+        if (body->len) {
+            PyObject *byte_object = PyBytes_FromStringAndSize((char *)(body->ptr), (Py_ssize_t)body->len);
+            if (PyFile_WriteObject(byte_object, request_binding->file_py, 0)) {
+                /* TODO: raise error */
+                goto done;
+            }
```
TingDaoK, line 129: The byte object will be converted to a string representation, https://github.com/python/cpython/blob/master/Objects/fileobject.c#L135. We need to come up with another approach if we want it to work well...
I guess we will have to create a FILE * from the fileno() given by Python in C land. Or @JonathanHenson, do you have any suggestion?

awscrt/s3.py
```diff
@@ -229,8 +229,6 @@ def __init__(
             region=None):
         assert isinstance(client, S3Client)
         assert isinstance(request, HttpRequest)
-        assert file is None or os.path.isfile(file)
```
JonathanHenson: I'm not sure we want the second half of the assertion. We might be implementing our own FILE* wrappers at some point that wouldn't be valid OS paths.

source/s3_meta_request.c
```diff
@@ -115,17 +120,27 @@ static void s_s3_request_on_body(
...
 
     /*************** GIL ACQUIRE ***************/
     PyGILState_STATE state;
     if (aws_py_gilstate_ensure(&state)) {
         return; /* Python has shut down. Nothing matters anymore, but don't crash */
     }
-    PyObject *result = PyObject_CallMethod(
-        request_binding->self_py, "_on_body", "(y#K)", (const char *)(body->ptr), (Py_ssize_t)body->len, range_start);
+    PyObject *result;
+    if (request_binding->file) {
```
JonathanHenson, line 139: these should likely be batched and sent something like once per second, instead of on each callback and be done outside of the context of the GIL

awscrt/s3.py
TingDaoK: Removed! Yeah, it should not be required to be a path

source/s3_meta_request.c
```diff
@@ -168,19 +242,20 @@ static void s_s3_request_on_finish(
...
+    result = PyObject_CallMethod(
+        request_binding->self_py,
+        "_on_finish",
+        "(iOy#)",
+        meta_request_result->error_code,
+        header_list,
+        (const char *)(error_body.buffer),
+        (Py_ssize_t)error_body.len);
+    if (header_list != Py_None) {
         Py_XDECREF(header_list);
```
graebm, line 258: to simplify the logic around whether or not you need to decref header_list, I'd suggest:
```
PyObject *result = NULL;
PyObject *header_list = NULL; // move this up to top of gil acquire block
...
PyObject_CallMethod(
    ...,
    header_list ? header_list : Py_NONE,
    ...);
...

done:
Py_XDECREF(header_list); // just one XDECREF(header_list)at end of gil block
```

source/s3_meta_request.c
```diff
@@ -310,6 +517,19 @@ PyObject *aws_py_s3_client_make_meta_request(PyObject *self, PyObject *args) {
     meta_request->self_py = py_s3_request;
     Py_INCREF(meta_request->self_py);
 
+    if (file_path) {
+        meta_request->file = fopen(file_path, type == AWS_S3_META_REQUEST_TYPE_GET_OBJECT ? "wb+" : "rb+");
+        if (type == AWS_S3_META_REQUEST_TYPE_PUT_OBJECT) {
+            meta_request->input_body = s_input_stream_new_from_open_file(allocator, meta_request->file, meta_request);
```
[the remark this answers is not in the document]
TingDaoK: We talked about it🙆‍♂️

awscrt/s3.py
```diff
@@ -141,6 +143,9 @@ def make_request(
                 AwsCredentials needed to sign an authenticated AWS request, for this request only.
                 If None is provided, the credential provider in the client will be used.
 
+            file (Optional[path]): Optional file path. If set, the C part will handle writing/reading from a file.
```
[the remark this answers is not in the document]
TingDaoK: yeah, from aws-c-s3, it will not invoke on_body callback for put_object. But still, for the sake of clarity, it's better to have separate file path

source/s3_meta_request.c
```diff
@@ -22,8 +25,24 @@ struct s3_meta_request_binding {
     /* Reference proxy to python self */
     PyObject *self_py;
 
+    /**
+     * FILE pointer. If set, it handles file operation from C land to reduce the cost
+     * passing chunks from C into python. One for recv/writing, the other for send/reading
+     **/
+    FILE *recv_file;
+    FILE *send_file;
```
graebm, line 33: why do we need `send_file`? doesn't `aws_input_stream_new_from_file()` already handle opening and closing a `FILE*`?
if you to open or close it in some special way, use `aws_input_stream_new_from_open_file()` instead

source/s3_meta_request.c
```diff
@@ -233,6 +324,122 @@ static void s_s3_request_on_shutdown(void *user_data) {
...
+    }
+
+    if (report_progress) {
+        /*************** GIL ACQUIRE ***************/
+        PyGILState_STATE state;
+        if (aws_py_gilstate_ensure(&state)) {
+            return AWS_OP_SUCCESS; /* Python has shut down. Nothing matters anymore, but don't crash */
+        }
+        PyObject *result =
+            PyObject_CallMethod(request_binding->self_py, "_on_progress", "(k)", request_binding->size_transferred);
```
graebm, line 362: `K` not `k`
https://docs.python.org/3/c-api/arg.html#c.Py_BuildValue

source/s3_meta_request.c
```diff
@@ -233,6 +324,122 @@ static void s_s3_request_on_shutdown(void *user_data) {
...
+        if (!result) {
+            PyErr_WriteUnraisable(request_binding->self_py);
```
graebm, line 364: probably want to raise an error and eventually return `AWS_OP_ERR`?
We can report errors in this particular callback

use this bad boy:
```
/**
 * Raise an AWS error corresponding to the current Python error.
 *
 * Prints the current Python error to stderr and clears the Python error indicator.
 * Finds an AWS error code corresponding to the current Python error (fallback is AWS_ERROR_UNKNOWN).
 * Invokes aws_raise_error() with that error code and always returns AWS_OP_ERR.
 *
 * The Python error indicator MUST be set and the GIL MUST be held when calling this function. */
int aws_py_raise_error(void);
```

source/s3_meta_request.c
```diff
@@ -242,26 +449,28 @@ PyObject *aws_py_s3_client_make_meta_request(PyObject *self, PyObject *args) {
     PyObject *s3_client_py = NULL;
     PyObject *http_request_py = NULL;
     int type;
-    PyObject *credential_provider_py;
+    PyObject *credential_provider_py = NULL;
+    const char *recv_filepath;
+    Py_ssize_t recv_filepath_len;
+    const char *send_filepath;
+    Py_ssize_t send_filepath_len;
```
graebm, line 456: trivial: since we don't actually need the len, you can get rid of `recv_filepath_len` and `send_filepath_len`. Use `z` instead of `z#` down in PyArg_ParseTuple()

source/s3_meta_request.c
```diff
@@ -310,6 +519,22 @@ PyObject *aws_py_s3_client_make_meta_request(PyObject *self, PyObject *args) {
     meta_request->self_py = py_s3_request;
     Py_INCREF(meta_request->self_py);
 
+    if (recv_filepath) {
+        meta_request->recv_file = fopen(recv_filepath, "wb+");
```
graebm, line 523: not checking for errors from fopen

source/s3_meta_request.c
```diff
@@ -310,6 +519,22 @@ PyObject *aws_py_s3_client_make_meta_request(PyObject *self, PyObject *args) {
...
+    if (send_filepath) {
+        meta_request->send_file = fopen(send_filepath, "rb+");
+        if (type == AWS_S3_META_REQUEST_TYPE_PUT_OBJECT) {
+            meta_request->input_body = s_input_stream_new_from_file(allocator, send_filepath, meta_request);
+            if (!meta_request->input_body) {
+                PyErr_SetAwsLastError();
+                goto error;
+            }
+            /* rewrite the input stream of the original request */
+            aws_http_message_set_body_stream(http_request, meta_request->input_body);
```
graebm, line 534: oh man, this is so weird... after this set_body_stream() call we have
`HttpRequest._binding.py_body_stream // still pointing to previous python body stream`
but
`HttpRequest._binding.native->body_stream // pointing to updated C body stream `

that's going to lead to something unexpected happening someday

Also, semi-related, we're not doing anything to keep http_request_py alive for the duration of the request. aws_input_stream still doesn't have refcounting. There's weird lifetime stuff with the python HttpRequest keeps the python InputStream alive. So if the python HttpRequest goes out of scope, the underlying aws_http_message will stay alive because s3 has a refcount on it, but the InputStream will get cleaned up which will destroy the non-ref-counted aws_input_stream

ANYWAY, here's my idea:
1) have the metarequest keep the python HttpRequest alive until the operation completes
2) if send_filepath is set, create a NEW aws_http_request that copies the headers/paths/etc of the one being passed in here, but uses a different aws_input_stream
3) if it's not set, you can just keep using the aws_http_request that got passed in

We'll end up copying the header strings, but it seems like a relatively small cost to pay

source/s3_meta_request.c
```diff
@@ -310,6 +519,22 @@ PyObject *aws_py_s3_client_make_meta_request(PyObject *self, PyObject *args) {
     meta_request->self_py = py_s3_request;
     Py_INCREF(meta_request->self_py);
 
+    if (recv_filepath) {
+        meta_request->recv_file = fopen(recv_filepath, "wb+");
+    }
+    if (send_filepath) {
+        meta_request->send_file = fopen(send_filepath, "rb+");
```
graebm, line 526: likewise, not checking results of fopen, but you probably don't need this code anyway...

source/s3_meta_request.c
```diff
@@ -120,75 +150,124 @@ static void s_s3_request_on_body(
...
+/* If the request has not finished, it will keep the request alive, until the finish callback invoked. So, we don't
+ * need to clean anything from this call */
 static void s_s3_request_on_finish(
     struct aws_s3_meta_request *meta_request,
     const struct aws_s3_meta_request_result *meta_request_result,
     void *user_data) {
     (void)meta_request;
     struct s3_meta_request_binding *request_binding = user_data;
+    if (request_binding->input_body) {
+        aws_input_stream_destroy(request_binding->input_body);
```
graebm, line 214: Noticed that this is the only place you're destroying the input_body.

Remember that a capsule can be in a partially-constructed state if aws_py_s3_client_make_meta_request() has an error.

You probably want to ALSO try and destroy it from s_destroy_if_ready() or the capsule destructor (you'd need to NULL out the variable here if you keep this destroy() in the on_finish)

I gotta go. But think about what will happen to ALL the stuff in your binding capsule if make_meta_request() fails at the last moment

source/s3_meta_request.c
TingDaoK: yeah, it's so weird... And it's so hard🤯

PR #218, opened 2021-01-02
cancel binding

TingDaoK:


test/test_s3.py
```diff
@@ -192,11 +203,110 @@ def test_put_object_file_object(self):
...
+        s3_client = s3_client_new(False, self.region, 5 * 1024 * 1024)
+        s3_request = s3_client.make_request(
+            request=http_request,
+            type=S3RequestType.PUT_OBJECT,
+            on_headers=self._on_request_headers)
+
+        if cancel_after_read:
+            read_futrue.result(self.timeout)
+
+        s3_request.cancel()
```
TingDaoK, line 289: I don't know about python...
Here is a deadlock I met.
Since we will hold lock to read from inputstream, https://github.com/awslabs/aws-c-s3/blob/main/source/s3_auto_ranged_put.c#L300-L309, which means the on_progress callback is invoked with the lock hold. That's dangerous....
And not sure how it's not back to the C land and keep excute python code in the same thread and try to canel the meta request, which also will acquire a lock.
BOOM, dead lock.

PR #220, opened 2021-01-07
numa-aware EventLoopGroup

graebm:
also, latest submodules


JonathanHenson (commented):
Could you add the cpu group detection stuff so that a user can detect whether or not they should pin?

JonathanHenson (approved):

PR #224, opened 2021-01-15
S3 follow up

TingDaoK:
*Issue #, if available:*

*Description of changes:*

- Remove the shutdown event for meta request.
- Invoke the finish call of request after the native client has cleaned up the request.
- Since once the request finishes, on further operations will happen on the request

### DIFF
--- a/source/s3_meta_request.c
+++ b/source/s3_meta_request.c
@@ -46,6 +46,9 @@ struct s3_meta_request_binding {
     uint64_t size_transferred;
     /* The time stamp when the progress reported */
     uint64_t last_sampled_time;
+
+    bool shutdown_called;
+    bool destructor_called;
 };
 
 struct aws_s3_meta_request *aws_py_get_s3_meta_request(PyObject *meta_request) {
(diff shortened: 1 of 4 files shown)


bretambrose (commented):
NM, real review was merged.

TingDaoK:
> Shouldn't all the printf's get removed?

Oh... I accidentally include the debug stuff to the last commit... Will remove... But the issue is our InputStream will throw an exception and trigger the python error handling, instead of the error handling we want to test here... So, not sure how to test the error handling issue properly.

source/s3_meta_request.c
```diff
@@ -23,6 +23,12 @@ struct s3_meta_request_binding {
     /* Reference to python object that reference to other related python object to keep it alive */
     PyObject *py_core;
 
+    /* If error headers included from finish call, store the python copy of the headers */
+    PyObject *error_headers;
+    /* If error body included from finish result, deep copy of the response body */
+    struct aws_byte_buf error_response_body;
+    int error_code;
```
graebm, line 30: probably simpler to make these changes in python rather than C, since this is a temporary workaround we'll be unwinding in the next few days. You could make the following changes in _S3RequestCore:
```
def _on_finish(self, error_code, error_headers, error_body):
    # temporary hack: delay on_done callback until shutdown
    self._done_error_code = error_code
    self._done_error_headers = error_headers
    self._done_error_body = error_body

    self._owner._binding = None # force S3Request's C parts to be cleaned up
    self._owner = None # allow S3Request to be cleaned up

def _on_shutdown(self):
    ... all the logic that used to be in _on_finish...
    ... the rest of the normal shutdown logic...

```

TingDaoK:
moved to https://github.com/awslabs/aws-crt-python/pull/232, which based just removed the workaround we have

PR #232, opened 2021-02-05
Metarequest refactor

TingDaoK:
*Issue #, if available:*

*Description of changes:*


awscrt/s3.py
```diff
@@ -274,17 +267,10 @@ def __init__(
 
         super().__init__()
 
-        self._on_headers_cb = on_headers
```
rccarper, line 277: Were these just unused?

rccarper (approved):
Looks good--if you update the submodule to v0.1.5, it should fix the cancel-test failure.

awscrt/s3.py
TingDaoK: `S3Request` don't need to hold a reference to them anymore, since the `_S3RequestCore` handles everything😀

PR #248, opened 2021-03-10
Updating submodules + use aws-lc for libcrypto

rccarper:
*Description of changes:*
Updating submodules to pull in latest aws-c-s3 fixes.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -121,16 +121,17 @@ def get_libcrypto_static_library(libcrypto_dir):
 
 
 class AwsLib:
-    def __init__(self, name, extra_cmake_args=[]):
+    def __init__(self, name, extra_cmake_args=[], link=True):
         self.name = name
         self.extra_cmake_args = extra_cmake_args
+        self.link = link
 
 
 # The extension depends on these libs.
 # They're built along with the extension, in the order listed.
 AWS_LIBS = []
 if sys.platform != 'darwin' and sys.platform != 'win32':
-    AWS_LIBS.append(AwsLib('aws-lc'))
+    AWS_LIBS.append(AwsLib('aws-lc', ['BUILD_LIBSSL=OFF'], False))
     AWS_LIBS.append(AwsLib('s2n'))
 AWS_LIBS.append(AwsLib('aws-c-common'))
 AWS_LIBS.append(AwsLib('aws-c-cal'))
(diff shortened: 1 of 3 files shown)


setup.py
```diff
@@ -197,14 +180,13 @@ def _build_dependency(self, aws_lib, libcrypto_paths):
...
-        if self.library_dirs:
-            cmake_args.append('-DCMAKE_LIBRARY_PATH="{}"'.format(';'.join(self.library_dirs)))
-        if libcrypto_paths:
-            cmake_args.append('-DLibCrypto_INCLUDE_DIR={}'.format(libcrypto_paths['include_dir']))
-            cmake_args.append('-DLibCrypto_STATIC_LIBRARY={}'.format(libcrypto_paths['static_library']))
+
+        # if self.include_dirs:
+        #     cmake_args.append('-DCMAKE_INCLUDE_PATH="{}"'.format(';'.join(self.include_dirs)))
+        # if self.library_dirs:
+        #     cmake_args.append('-DCMAKE_LIBRARY_PATH="{}"'.format(';'.join(self.library_dirs)))
```
graebm, line 189: delete

setup.py
```diff
@@ -148,6 +130,7 @@ def __init__(self, name, extra_cmake_args=[]):
 # They're built along with the extension, in the order listed.
 AWS_LIBS = []
 if sys.platform != 'darwin' and sys.platform != 'win32':
+    AWS_LIBS.append(AwsLib('aws-lc'))
```
graebm, line 133: It's not merged yet, but there's an outstanding PR for aws-lc that will let us turn off building the parts of aws-lc that we don't need: https://github.com/awslabs/aws-lc/pull/105

Anyway, add this extra cmake arg to aws-lc. When that PR gets merged it will eventually speed us up

`AWS_LIBS.append(AwsLib('aws-lc', ['BUILD_LIBSSL=OFF']))`

setup.py
```diff
@@ -256,7 +234,7 @@ def awscrt_ext():
     extra_link_args = os.environ.get('LDFLAGS', '').split()
     extra_objects = []
 
-    libraries = [x.name for x in AWS_LIBS]
+    libraries = [x.name for x in AWS_LIBS if x.name != 'aws-lc']
```
graebm, line 237: is this necessary? if we're not directly linking aws-lc, why are we directly linking s2n?

If we do need this, let's make it a little more data-driven and give a `link` property to each `AwsLib`. So we'd create the lib like `AwsLib('s2n', link=False)`, and then here we'd be like:
libraries = [x.name for x in AWS_LIBS if x.link]

graebm (commented):
Please rename this PR to something like "use aws-lc for libcrypto".
"Updating submodules" is so innocuous

setup.py
justinboswell: +1

setup.py
justinboswell: Works for me, I think the even better would be to have `self.link=self.name` by default, and then set it to ':libcrypto.a' for aws-lc.

graebm: I'm confused...
NOW I see that we didn't include "aws-lc" in the list of libraries to link because we want to link to "libcrypto.a" and the original code would have gotten us "libaws-lc.a".

But with the code as it's written here, we don't include anything about libcrypto in the `libraries` list. Is it somehow getting pulled in transitively through s2n and aws-c-cal? Or is it magic that's somehow letting this all pass?

graebm: figured it out
we were doing some weird stuff with `libraries` in most code-paths
cleaned it up and pushed

graebm (approved):

PR #266, opened 2021-04-15
Fix native memory leaks

graebm:
All tests now check for native memory leaks. Previously, we only checked whether python classes inheriting from `NativeResource` were leaked.

Tests must now be run with the `AWS_CRT_MEMORY_TRACING=2` environment variable set.

The following leaks were found and fixed:
- `awscrt.http.HttpRequest` was leaking its headers
- `awscrt.eventstream.rpc.ClientConnection` was leaking its native part
- `awscrt.auth.AwsSigningConfig` was leaking its native part
- each call to `awscrt.auth.aws_sign_request()` was leaking native native data
- `eventstream.rpc` was leaking incoming headers

### DIFF
--- a/awscrt/_test.py
+++ b/awscrt/_test.py
@@ -134,7 +134,7 @@ def check_for_leaks(*, timeout_sec=10.0):
                 if isinstance(r, types.FrameType):
                     frameinfo = inspect.getframeinfo(r)
                     our_fault = (frameinfo.filename.endswith('_weakrefset.py') or
-                                 frameinfo.filename.endswith('test/__init__.py'))
+                                 frameinfo.filename.endswith('awscrt/_test.py'))
                     if our_fault:
                         continue
 
--- a/awscrt/_test.py
+++ b/awscrt/_test.py
@@ -1,0 +1,157 @@
+"""
+Private utilities for testing
+"""
+import _awscrt
+from awscrt import NativeResource
+import gc
+import inspect
+import os
+import sys
+import time
+import types
+
+
+def native_memory_usage() -> int:
+    """
+    Returns number of bytes currently allocated by awscrt's native code.
+
+    `AWS_CRT_MEMORY_TRACING `environment variable must be set before module
+    is loaded, or 0 will always be returned. Legal values are:
+
+    *   `AWS_CRT_MEMORY_TRACING=0`: No tracing
+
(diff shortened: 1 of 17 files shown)


awscrt/_test.py
```diff
@@ -0,0 +1,157 @@
...
+        timeout_sec (float): Number of seconds to wait before a timeout exception is raised.
+            By default the wait is unbounded.
+
+    Returns:
+        bool: Returns whether threads could be joined before the timeout.
+    """
+    return _awscrt.thread_join_all_managed(timeout_sec)
+
+
+def check_for_leaks(*, timeout_sec=10.0):
```
graebm, line 65: The contents of this function used to live in `tests.__init__.py`

awscrt/_test.py
```diff
@@ -0,0 +1,157 @@
...
+    *   `AWS_CRT_MEMORY_TRACING=0`: No tracing
+
+    *   `AWS_CRT_MEMORY_TRACING=1`: Only track allocation sizes and total allocated
+
+    *   `AWS_CRT_MEMORY_TRACING=2`: Capture callstacks for each allocation
+    """
+    return _awscrt.native_memory_dump()
+
+
+def join_all_native_threads(*, timeout_sec: float = -1.0) -> bool:
```
graebm, line 48: Hiding some of these new functions in a private `_test.py` file so that we're free to change them going forward. If we have a need to make them public we can move them to `common.py`

DavidOgunsAWS (approved):

source/event_stream_headers.c
```diff
@@ -255,15 +255,17 @@ PyObject *aws_py_event_stream_python_headers_create(
...
-            header->header_name,
-            header->header_name_len,
-            /* NOTE: if s_create_python_header_value() returns NULL,
-             * then PyObject_CallFunction() fails too, which is convenient */
-            s_create_python_header_value(header),
-            header->header_value_type);
+        PyObject *tuple_py =
+            Py_BuildValue("(s#Oi)", header->header_name, header->header_name_len, value_py, header->header_value_type);
+
+        Py_DECREF(value_py);
```
justinboswell, line 267: Since you commented the new reference, should probably comment this decref

justinboswell (approved):

PR #268, opened 2021-04-19
Metadata/Version/Quality of Life updates

### LINKED ISSUE #257: Feature Request: Add __version__ to awscrt
Most well formed Python packages will provide a dunder version attribute on their top level package name. This provides a uniform way to get versioning metadata without requiring `pkg_resources` look ups. There's a number of ways that have been used to integrate this value into your setup.py, so I've provided a few examples for ideas. This is minor quality of life fix, but brings crt more in line with the rest of the ecosystem.

Examples:
* [Pip \_\_init\_\_.py](https://github.com/pypa/pip/blob/main/src/pip/__init__.py#L3), [Pip setup.py](https://github.com/pypa/pip/blob/main/setup.py#L30)
* [Requests \_\_version\_\_.py](https://github.com/psf/requests/blob/master/requests/__version__.py#L8), [Requests setup.py](https://github.com/psf/requests/blob/master/setup.py#L68)
* [urllib3 _version.py](https://github.com/urllib3/urllib3/blob/main/src/urllib3/_version.py#L2), [urllib3 setup.py](https://github.com/urllib3/urllib3/blob/main/setup.py#L40-L44)

nateprewitt:
## Overview
This PR should resolve #257 and address a couple other low hanging fixes for the AWS CRT Python bindings.

### Add __version__ attribute

This piece addresses #257 by adding moving the version specifier for this package out of the `setup.py` and into a more accessible place in the top level `__init__.py` file. This will allow easy access on a static attribute (e.g. `awscrt.__version__`) bringing this package more inline with Python standards.

### Add a long_description

While I was fixing the version, I noticed we're not uploading any useful readme information to PyPI. The package is currently just a [blank page](https://pypi.org/project/awscrt/). I added the readme as the `long_description` which is standard practice to provide general information to users on PyPI.

**Note this syntax requires `wheel>=31.0` to work correctly. I couldn't find release dependencies but that needs to be verified before merging.**

### Add a version badge to README
Added basic version badge to the README that updates with each release so anyone can see the latest version from their already downloaded copy.

### Fixed License Specifier

The Trove Classifier for the license was also used in the license field when the package was set up which generates odd metadata on PyPI. I moved it to the expected plain text specifier so it displays correctly now.

## Testing

I ran the `update_version.py` script and the `test-version-exists` script to test. I wasn't able to find other testing infrastructure but happy to do more if you have suggestions.


graebm (approved):

graebm:
Closing this PR. It's duplicated here: https://github.com/awslabs/aws-crt-python/pull/286
I duplicated this branch so that our CI would run (it doesn't run on PRs from forks)
Then it needed some stupid code formatting so I started pushing more changes

PR #269, opened 2021-04-19
close input stream when meta request finishes

TingDaoK:
*Issue #, if available:*

- During the on_done callback, which will be invoked as meta request finishes, the caller may touch the input stream file. Eg: `aws s3 mv foot.txt s3://bucket`. Windows will block it.

*Description of changes:*

- We don't need the input stream as meta request finishes, close it earlier 
- Copy/Paste a file helper from https://github.com/boto/s3transfer/blob/develop/tests/__init__.py#L153. Since we may need it later

### DIFF
--- a/test/__init__.py
+++ b/test/__init__.py
@@ -8,6 +8,10 @@ import sys
 import time
 import types
 import unittest
+import tempfile
+import math
+import os
+import shutil
 
 TIMEOUT = 10.0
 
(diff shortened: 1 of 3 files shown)


test/__init__.py
```diff
@@ -71,3 +75,53 @@ def _printobj(prefix, obj):
                         _printobj('  -', r)
 
         self.assertEqual(0, len(NativeResource._living))
+
+
+class FileCreator(object):
```
graebm, line 80: style: Just move this helper class to test_s3.py where you're using it

`__init__.py` isn't a great place for stuff because it gets included in EVERYTHING below it. In retrospect, I wish all the `__init__.py` files were empty

test/test_s3.py
```diff
@@ -88,6 +92,11 @@ def setUp(self):
 
         self.put_body_stream = None
 
+        self.files = FileCreator()
+
+    def tearDown(self):
+        self.files.remove_all()
```
graebm, line 98: need to call `super().tearDown()` afterwards
`NativeResourceTest.tearDown()` is where we check for leaks

graebm (approved):
fix & ship

PR #283, opened 2021-05-11
Crc

ilevyor:
*Issue #, if available:*

*Description of changes:*


source/crc.c
```diff
@@ -0,0 +1,57 @@
...
+ */
+
+#include "checksums.h"
+
+#include "aws/checksums/crc.h"
+#include "aws/common/byte_buf.h"
+
+PyObject *aws_py_checksums_crc32(PyObject *self, PyObject *args) {
+    (void)self;
+    // (void)args;
```
rccarper, line 13: Trivial: remove.

source/crc.c
```diff
@@ -0,0 +1,57 @@
...
+    struct aws_byte_buf input = aws_byte_buf_from_array(input_buf.buf, input_buf.len);
+
+    uint32_t crc_res = aws_checksums_crc32(input.buffer, input.len, previousCrc32);
+
+    return PyLong_FromUnsignedLong(crc_res);
+}
+
+PyObject *aws_py_checksums_crc32c(PyObject *self, PyObject *args) {
+    (void)self;
+    // (void)args;
```
rccarper, line 37: Trival: remove.

source/crypto.c
```diff
@@ -29,6 +29,29 @@ static void s_hmac_destructor(PyObject *hmac_capsule) {
...
+
+    struct aws_hash *sha1 = aws_sha1_new(allocator);
+
+    if (!sha1) {
+        return PyErr_AwsLastError();
+    }
+
+    PyObject *capsule = PyCapsule_New(sha1, s_capsule_name_hash, s_hash_destructor);
+
+    /* Why do we return NULL instead of PyErr */
```
rccarper, line 46: Is this still a question we have?  Can we remove this comment?

source/crc.c
```diff
@@ -0,0 +1,57 @@
...
+    PyObject *py_previousCrc32;
+
+    if (!PyArg_ParseTuple(args, "s*O", &input_buf, &py_previousCrc32)) {
+        return PyErr_AwsLastError();
+    }
+
+    uint32_t previousCrc32 = PyLong_AsUnsignedLong(py_previousCrc32);
+
+    if (previousCrc32 == (uint32_t)-1 && PyErr_Occurred()) {
+        return NULL;
```
rccarper, line 25: Should we return PyErr_AwsLastError here (like above) instead of NULL?

source/crc.c
```diff
@@ -0,0 +1,57 @@
...
+    }
+
+    struct aws_byte_buf input = aws_byte_buf_from_array(input_buf.buf, input_buf.len);
```
rccarper, line 28: Trivial: Are these conversions to aws_byte_buf redundant?  (It looks like it just copies the pointer/len directly into the byte buffer structure, and then we re-feed the same values into aws_checksums_crc32)

rccarper (changes requested):

rccarper:
Also just leaving a note so that we don't forget to change aws-c-cal to a released version before merging. :)

source/crc.c
ilevyor: No, this is correct. I still don't quite understand why. @graebm python will automatically add a useful error in the case of NULL. Not sure exactly what the difference is between the two situations

source/crypto.c
ilevyor: trivial remove. Same question you had in the above comment.

source/crc.c
TingDaoK: So, functions from python C extension will raise the error for python. But the function from our lib will not. In that case, we will need to raise the Python Error from our aws last error, which is `PyErr_AwsLastError` for. EG: `PyArg_ParseTuple` is a function from python C extension, you can safely return NULL at line 18.

ilevyor: oh that makes sense, I'll change 
>        return PyErr_AwsLastError();
to NULL then to remove the confusion.

source/crc.c
```diff
@@ -0,0 +1,57 @@
...
+
+#include "checksums.h"
+
+#include "aws/checksums/crc.h"
+#include "aws/common/byte_buf.h"
+
+PyObject *aws_py_checksums_crc32(PyObject *self, PyObject *args) {
+    (void)self;
+
+    Py_buffer input_buf;
```
graebm: Py_buffer needs to be freed when you're done with it. See how cpython did it:
https://github.com/python/cpython/blob/ff230153fd2b1d0576df21ae6c29f0f2fdb0244f/Modules/clinic/zlibmodule.c.h#L775

source/crc.c
```diff
@@ -0,0 +1,53 @@
...
+}
+
+PyObject *aws_py_checksums_crc32c(PyObject *self, PyObject *args) {
+    (void)self;
+
+    Py_buffer input_buf;
+    PyObject *py_previousCrc32c;
+
+    if (!PyArg_ParseTuple(args, "s*O", &input_buf, &py_previousCrc32c)) {
+        return PyErr_AwsLastError();
```
TingDaoK, line 39: Same as above.

source/crypto.c
```diff
@@ -56,7 +85,15 @@ PyObject *aws_py_md5_new(PyObject *self, PyObject *args) {
         return PyErr_AwsLastError();
     }
 
-    return PyCapsule_New(md5, s_capsule_name_hash, s_hash_destructor);
+    PyObject *capsule = PyCapsule_New(md5, s_capsule_name_hash, s_hash_destructor);
+
+    /* Why do we return NULL instead of PyErr */
```
TingDaoK, line 90: Remove this comment. Check the [doc](https://docs.python.org/3/c-api/capsule.html#c.PyCapsule_New) here.
> On failure, set an exception and return NULL.

Basically those functions will set an exception on failure.

source/crc.c
```diff
@@ -0,0 +1,57 @@
...
+    if (previousCrc32 == (uint32_t)-1 && PyErr_Occurred()) {
+        return NULL;
+    }
+
+    // struct aws_byte_buf input = aws_byte_buf_from_array(input_buf.buf, input_buf.len);
+
+    // uint32_t crc_res = aws_checksums_crc32(input.buffer, input.len, previousCrc32);
+
+    uint32_t crc_res = aws_checksums_crc32(input_buf.buf, input_buf.len, previousCrc32);
+
```
graebm, line 28: Py_buffer.buf isn't necessarily a contiguous block of memory. [Docs](https://docs.python.org/3/c-api/buffer.html#c.Py_buffer):
> For contiguous arrays, this is the length of the underlying memory block. For non-contiguous arrays, it is the length that the logical structure would have if it were copied to a contiguous representation

Python's built-in CRC function [checks if the buffer is contiguous](https://github.com/python/cpython/blob/ff230153fd2b1d0576df21ae6c29f0f2fdb0244f/Modules/clinic/zlibmodule.c.h#L760), we should too.

source/crc.c
ilevyor: corrected

source/crypto.c
```diff
@@ -41,7 +63,15 @@ PyObject *aws_py_sha256_new(PyObject *self, PyObject *args) {
...
     }
 
-    return PyCapsule_New(sha256, s_capsule_name_hash, s_hash_destructor);
+    PyObject *capsule = PyCapsule_New(sha256, s_capsule_name_hash, s_hash_destructor);
+
+    /* Why do we return NULL instead of PyErr */
+    if (capsule == NULL) {
+        aws_hash_destroy(sha256);
+        return NULL;
+    }
```
graebm: Read this: https://docs.python.org/3/extending/extending.html#intermezzo-errors-and-exceptions

then remove comment

the call to `PyErr_AwsLastError()` doesn't actually return an exception. It sets an exception in a global static variable and just returns NULL, which is how python-c-api functions indicate that an error occurred (telling python to look at hte global static variable for details on the error)

source/crypto.c
```diff
@@ -56,7 +86,15 @@ PyObject *aws_py_md5_new(PyObject *self, PyObject *args) {
         return PyErr_AwsLastError();
     }
 
-    return PyCapsule_New(md5, s_capsule_name_hash, s_hash_destructor);
+    PyObject *capsule = PyCapsule_New(md5, s_capsule_name_hash, s_hash_destructor);
+
+    /* Why do we return NULL instead of PyErr */
```
graebm, line 90: remove comment

graebm (changes requested):
read this: https://docs.python.org/3/extending/extending.html#intermezzo-errors-and-exceptions

source/crc.c
ilevyor: Should I just set the argument to bad argument and return null if that's the case, like what they do in the example?

source/crc.c
```diff
@@ -0,0 +1,65 @@
...
+#include "checksums.h"
+
+#include "aws/checksums/crc.h"
+#include "aws/common/byte_buf.h"
+
+PyObject *aws_py_checksums_crc32(PyObject *self, PyObject *args) {
+    (void)self;
+
+    Py_buffer input;
```
graebm, line 14: I made a small comment about this in the last review, but it got vanished so I'll make it again:
Py_buffer needs to be cleaned up if `PyArg_ParseTuple()` succeeds.

Lots of things in C that get created successfully, need to be manually "cleaned up" or "released" when you're done with them. Things like `int` don't need to be cleaned up because they're just some bytes on the stack. Things like `PyObject *` that we get from `PyArg_ParseTuple()` don't need to be cleaned up because they're just passed into this function by pointer (it's a "borrow" in Rust terms). But `Py_buffer` is different. Notice that it's by-value and not by-pointer. If `PyArg_ParseTuple()` succeeds, then it has "initialized" `Py_buffer` in a way that it needs to be cleaned up when we're done with it. How does a beginner learn this subtle detail? They read a [lot](https://docs.python.org/3/extending/extending.html) [of](https://docs.python.org/3/c-api/arg.html#strings-and-buffers) [docs](https://docs.python.org/3/c-api/buffer.html) 😬. It needs to be cleaned up whether `aws_py_checksums_crc32()` succeeds or fails. And this is why `goto` is popular in C. `goto` is not popular in any other language because in any other language you can make cleanup happen automatically as part of a type's destructor. Yeah ... there's a reason people don't write C anymore.

You're going to want the function to look something like this (incoming pseudocode!)
```
    Py_buffer input; /* Py_buffers must be released after successful PyArg_ParseTuple() calls */
    if (!PyArg_ParseTuple(...))
        return NULL

    /* From hereon, we need to clean up before returning */
    PyObject *py_result = NULL;

    ...
    If (something went wrong like the contiguous check, or the int checks)
        goto done;
    ...

   
    py_result = PyLong_FromUnsignedLong(crc_res);
done:
    PyBuffer_Release(&payload_buf);
    return py_result;
```

source/crc.c
graebm: you don't literally want to `return NULL;` because we need to cleanup the buffer before returning, so we'll probably use a `goto`. But yes, in the end, you just want NULL to get returned

ilevyor: why goto over calling a cleanup function?

graebm: If you use a goto you can put all your cleanup code in one place.
If you don't use a goto, you need to copy/paste the cleanup code everywhere you do a return... and then in the future the function changes so that there's more cleanup and someone needs to find every return and copy/paste the extra code into all the different places cleanup happens ... and they miss something somewhere

cleanup in C is easier to manage with a goto

source/crc.c
```diff
@@ -0,0 +1,119 @@
...
+            signed_val = aws_checksums_crc32(buf, (unsigned int)len, signed_val);
+        Py_END_ALLOW_THREADS
+        /* clang-format on */
+    } else {
+        signed_val = aws_checksums_crc32(input.buf, (unsigned int)input.len, signed_val);
+    }
+    if (input.obj) {
+        PyBuffer_Release(&input);
+    }
+    /* why the bitwise and? */
```
graebm, line 63: remove this comment

source/crc.c
```diff
@@ -0,0 +1,119 @@
...
+        Py_ssize_t len = input.len;
+
+        /* clang-format off */
+        Py_BEGIN_ALLOW_THREADS
+            /* Avoid truncation of length for very large buffers. crc32() takes
+               length as an unsigned int, which may be narrower than Py_ssize_t. */
+            while ((size_t)len > UINT_MAX) {
+                signed_val = aws_checksums_crc32(buf, UINT_MAX, signed_val);
+                buf += (size_t)UINT_MAX;
+                len -= (size_t)UINT_MAX;
```
graebm, line 52: aws_checksums_crc32() takes length as an `int`, not a `uint`
~~length as an unsigned int,~~ -> length as a signed int
~~UINT_MAX~~ -> INT_MAX

add a test that hits this branch. It's not correct at the moment, and definitely worth testing given the complexity

I'm actually pretty spooked why the compiler isn't complaining about this

graebm: for the test, it might very well fail on a computer with insufficient memory, so the code would be something like:
```
# stress the internal logic that handles buffers larger than C's INT_MAX
try:
    INT_MAX = 2**31 - 1
    giant_buffer = bytes(INT_MAX + 5)
except MemoryError:
    raise unittest.SkipTest('Machine can't allocate giant buffer for giant buffer test')

val = crc32(giant_buffer)
```

source/crc.c
```diff
@@ -0,0 +1,119 @@
...
+    if (!PyBuffer_IsContiguous(&input, 'C')) {
+        PyErr_SetString(PyExc_ValueError, "input must be contiguous buffer");
+        return NULL;
+    }
+
+    int signed_val = previousCrc32c;
+
+    /* Releasing the GIL for very small buffers is inefficient
+       and may lower performance */
+    if (input.len > 1024 * 5) {
```
graebm, line 97: Given how complex these 2 functions have gotten, I would create a single implementation, that only differs on which crc function gets called under the hood. crc32 and crc32c have identical function signatures, so you could just pass the crc functions as an argument to the common implementation

source/crypto.c
```diff
@@ -78,8 +115,22 @@ PyObject *aws_py_hash_update(PyObject *self, PyObject *args) {
     struct aws_byte_cursor to_hash_cursor;
     to_hash_cursor = aws_byte_cursor_from_array(to_hash_c_str, to_hash_len);
 
-    if (aws_hash_update(hash, &to_hash_cursor)) {
-        return PyErr_AwsLastError();
+    if (to_hash_len > 1024 * 5) {
```
graebm, line 118: add comment
```
    /* Releasing the GIL for very small buffers is inefficient
       and may lower performance */
```

test/test_crypto.py
```diff
@@ -0,0 +1,79 @@
...
+    def test_sha1_iterated(self):
+        h = Hash.sha1_new()
+        h.update('a')
+        h.update('b')
+        h.update('c')
+        digest = h.digest()
+        expected = b'\xa9\x99\x3e\x36\x47\x06\x81\x6a\xba\x3e\x25\x71\x78\x50\xc2\x6c\x9c\xd0\xd8\x9d'
+        self.assertEqual(expected, digest)
+
+    def test_md5_empty(self):
```
graebm, line 55: add a `test_sha1_large_buffer()` test so you hit the branch that releases the GIL
you don't need to do an INT_MAX buffer, just something larger than the 5KB that trigger releasing the GIL

source/crypto.c
graebm: trivial: still need to add this comment

test/test_checksums.py
```diff
@@ -0,0 +1,100 @@
...
+
+    def test_crc32_values_iterated(self):
+        output = 0
+        for i in range(32):
+            output = checksums.crc32(chr(i), output)
+        expected = 0x91267E8A
+        self.assertEqual(expected, output)
+
+# stress test gil optimization for 32 bit architecture which cannot handle hug buffer
+    def test_crc32_large_buffer(self):
```
graebm, line 37: trivial: put the comment *inside* the function. Python's weird like that. But that's why the formatter indented the comment in a weird way

also: ~~hug~~ -> huge

test/test_checksums.py
```diff
@@ -0,0 +1,100 @@
...
+        try:
+            giant_buffer = bytes(25 * 2**20)
+        except (MemoryError, OverflowError):
+            raise unittest.SkipTest('Machine cant allocate giant buffer for giant buffer test')
```
graebm, line 41: trivial: remove error handling here. any machine should be able to do a buffer this size

test/test_checksums.py
```diff
@@ -0,0 +1,100 @@
...
+        val = checksums.crc32(giant_buffer)
+        self.assertEqual(0x72103906, val)
+
+# stress the internal logic that handles buffers larger than C's INT_MAX
+    def test_crc32_huge_buffer(self):
```
graebm, line 46: trivial: comment inside function

test/test_checksums.py
```diff
@@ -0,0 +1,100 @@
...
+            output = checksums.crc32c(chr(i), output)
+        expected = 0x46DD794E
+        self.assertEqual(expected, output)
+
+# stress test gil optimization for 32 bit architecture which cannot handle hug buffer
+    def test_crc32c_large_buffer(self):
+        try:
+            giant_buffer = bytes(25 * 2**20)
+        except (MemoryError, OverflowError):
+            raise unittest.SkipTest('Machine cant allocate giant buffer for giant buffer test')
```
graebm, line 84: trivial: likewise: comment inside function. remove error handling on this function

graebm (approved):
Fix and ship!

test/test_checksums.py
ilevyor: when should python comments be strings?

rccarper (approved):

PR #288, opened 2021-05-13
Fix possible Py_buffer leak.

graebm:
While pointing out a memory-leak in another PR, I noticed I'd made the same mistake a few months back 😅

### DIFF
--- a/source/event_stream_rpc_client_continuation.c
+++ b/source/event_stream_rpc_client_continuation.c
@@ -180,7 +180,7 @@ PyObject *aws_py_event_stream_rpc_client_continuation_activate(PyObject *self, P
     bool success = false;
     struct aws_array_list headers;
     AWS_ZERO_STRUCT(headers);
-    bool set_self_py = false;
+    bool self_py_set = false;
     Py_INCREF(on_flush_py); /* Keep completion callback alive until it fires */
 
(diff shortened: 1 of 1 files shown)


source/event_stream_rpc_client_continuation.c
```diff
@@ -175,25 +175,27 @@ PyObject *aws_py_event_stream_rpc_client_continuation_activate(PyObject *self, P
         return NULL;
     }
 
+    /* From hereon, we need to clean up if errors occur */
+
+    bool success = false;
```
justinboswell, line 180: This is never used

justinboswell: Unless this is an artistic commentary on software. In which case, leave it.

justinboswell (approved):

bretambrose (approved):

source/event_stream_rpc_client_continuation.c
graebm: It's used:
```
if (success) {
        Py_RETURN_NONE;
    }

    /* failed */
    Py_DECREF(on_flush_py);
    if (self_py_set) {
        Py_CLEAR(continuation->self_py);
    }

    return NULL;
    ```

PR #292, opened 2021-05-24
Fixing null pointer bug when connection setup error occurred.

DavidOgunsAWS:
*Description of changes:*
* When a connection setup error has occurred, native connection is null so acquire() shouldn't be invoked

### DIFF
--- a/source/event_stream_rpc_client_connection.c
+++ b/source/event_stream_rpc_client_connection.c
@@ -128,10 +128,12 @@ static void s_on_connection_setup(
     struct aws_event_stream_rpc_client_connection *native,
     int error_code,
     void *user_data) {
-
     struct connection_binding *connection = user_data;
     connection->native = native;
-    aws_event_stream_rpc_client_connection_acquire(connection->native);
+
+    if (native) {
+        aws_event_stream_rpc_client_connection_acquire(connection->native);
+    }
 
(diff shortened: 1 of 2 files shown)


source/event_stream_rpc_client_connection.c
```diff
@@ -152,8 +154,9 @@ static void s_on_connection_setup(
             (void *)connection->native);
 
         PyErr_WriteUnraisable(connection->self_py);
-
-        aws_event_stream_rpc_client_connection_close(connection->native, AWS_ERROR_CRT_CALLBACK_EXCEPTION);
+        if (native) {
+            aws_event_stream_rpc_client_connection_close(connection->native, AWS_ERROR_CRT_CALLBACK_EXCEPTION);
```
graebm, line 158: 😅
another nice catch

test/test_eventstream_rpc.py
```diff
@@ -92,9 +92,7 @@ def on_continuation_closed(self, **kwargs):
         self.record.close_call.set()
 
 
-@skipUnless(RUN_LOCALHOST_TESTS, "Skipping until we have permanent echo server")
-class TestClient(NativeResourceTest):
-
+class FailureClientTests(NativeResourceTest):
```
graebm, line 95: Add comment about how these tests can run without a server
I had to think about why you split these up into 2 classes

graebm (approved):

PR #295, opened 2021-06-02
Sigv4a

bretambrose:
* Sigv4a implementation


awscrt/auth.py
```diff
@@ -401,9 +404,13 @@ class AwsSigningConfig(NativeResource):
             computed from the signable.
 
         credentials_provider (AwsCredentialsProvider): Credentials provider
-            to fetch signing credentials with.
+            to fetch signing credentials with. If the signing algorithm is
+            V4_ASYMMETRIC, ecc-based credentials will be derived from the
```
graebm, line 408: Turn this into a link like so (note the backticks):
```
`:attr:AwsSigningAlgorithm.V4_ASYMMETRIC`
```

graebm: also: capitalize ECC

awscrt/auth.py
```diff
@@ -550,12 +557,20 @@ def signature_type(self):
 
     @property
     def credentials_provider(self):
-        """AwsCredentialsProvider: Credentials provider to fetch signing credentials with"""
+        """
+        AwsCredentialsProvider: Credentials provider to fetch signing credentials with.
+        If the signing algorithm is V4_ASYMMETRIC, ecc-based credentials will be derived
```
graebm, line 562: likewise, turn V4_ASYMMETRIC into link and capitalize ECC

awscrt/auth.py
```diff
@@ -401,9 +404,13 @@ class AwsSigningConfig(NativeResource):
...
 
         credentials_provider (AwsCredentialsProvider): Credentials provider
-            to fetch signing credentials with.
+            to fetch signing credentials with. If the signing algorithm is
+            V4_ASYMMETRIC, ecc-based credentials will be derived from the
+            fetched credentials.
 
-        region (str): The region to sign against.
+        region (str): If signing algorithm is V4, the region to sign against.
+            If signing algorithm is V4_ASYMMETRIC, the value of the
```
graebm, line 412: links for V4 and V4_ASYMMETRIC

awscrt/auth.py
```diff
@@ -401,9 +404,13 @@ class AwsSigningConfig(NativeResource):
...
+            X-amzn-region-set header (added in signing).
```
graebm, line 413: trivial: quotes around "X-amzn-region-set" we do that when referencing other headers/query-params

awscrt/auth.py
```diff
@@ -550,12 +557,20 @@ def signature_type(self):
...
+        """
         return _awscrt.signing_config_get_credentials_provider(self._binding)
 
     @property
     def region(self):
-        """str: The region to sign against"""
+        """
+        str: If signing algorithm is V4, the region to sign against.
+        If signing algorithm is V4_ASYMMETRIC, the value of the
+        X-amzn-region-set header (added in signing).
```
graebm, line 572: same as above, links and quotes around header name

source/auth_signer.c
```diff
@@ -105,6 +105,7 @@ PyObject *aws_py_sign_request_aws(PyObject *self, PyObject *args) {
 
     async_data->signable = aws_signable_new_http_request(aws_py_get_allocator(), http_request);
     if (!async_data->signable) {
+        PyErr_SetAwsLastError();
```
graebm, line 108: 😱
great catch

graebm (commented):
re-generate the docs too, if you're able

graebm (approved):
fix & ship

TingDaoK:
have no `aws_verify_sigv4a_signing` binding for python... Do we need it?

source/auth_signer.c
bretambrose: git blame says you added this!

bretambrose:
> have no `aws_verify_sigv4a_signing` binding for python... Do we need it?



> have no `aws_verify_sigv4a_signing` binding for python... Do we need it?

we'd need it if we checked the actual signature value in the test, but we don't

PR #317, opened 2021-10-07
Python Abi3 

TingDaoK:
*Issue #, if available:*

*Description of changes:*


TingDaoK:
TO support ABI3, we need to remove all those function we used that is NOT part of the [limited api](https://docs.python.org/3/c-api/stable.html#contents-of-limited-api). Comparing to the complexity it will have to just add the new version of python to our pipeline seems not really worth.

More importantly, as we can see from the performance related limitation with the stable ABI [here](https://docs.python.org/3/c-api/stable.html#limited-api-scope-and-performance), which makes us more concerning about supporting abi3. 

Closing the PR for now, maybe later when Python introduces a version that will ensure stable ABI and no performance impact after that, we can pick that version.

TingDaoK:
It's supported from https://github.com/awslabs/aws-crt-python/pull/479

PR #320, opened 2021-12-04
Tests use env vars so they're easier to run locally

graebm:
ISSUE: Previously, if a test needed user-specific configuration to run, it would pull things like keys and certs from AWS secretsmanager using boto3. This meant developers that wanted to run the tests locally needed to configure their AWS account exactly like the test expected (inconvenient) or use the dev-team's credentials (kinda dangerous).

DESCRIPTION OF CHANGES: Use environment variables to pass things like certs and key. When CI runs, it still pulls the values from secretsmanager, and sets env-vars accordingly. But ordinary developers can just use env-vars.

### DIFF
--- a/.builder/actions/aws_crt_python.py
+++ b/.builder/actions/aws_crt_python.py
@@ -1,8 +1,11 @@
 
 import Builder
 import argparse
-import os
+import os.path
+import pathlib
+import subprocess
 import sys
+import tempfile
 
 
 class InstallPythonReqs(Builder.Action):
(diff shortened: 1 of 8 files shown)


sbSteveK (approved):

.builder/actions/aws_crt_python.py
```diff
@@ -27,6 +30,45 @@ def run(self, env):
...
+        setenv_from_secret('AWS_TEST_IOT_MQTT_ENDPOINT', 'unit-test/endpoint')
+
+        setenv_tmpfile_from_secret('AWS_TEST_TLS_CERT_PATH', 'unit-test/certificate', 'certificate.pem')
+        setenv_tmpfile_from_secret('AWS_TEST_TLS_KEY_PATH', 'unit-test/privatekey', 'privatekey.pem')
+
+        # tests must run with leak detection turned on
+        env.shell.setenv('AWS_CRT_MEMORY_TRACING', '2')
+
+        # enable S3 tests
+        env.shell.setenv('AWS_TEST_S3', '1')
```
TingDaoK, line 69: Just thoughts: Is it worth to put those environment variables to README?
- it will be easier for people to know about. 
- But, we do have warnings when we skip tests about the env variables. Hmmm.

TingDaoK (approved):
Great improvement❤️

PR #323, opened 2021-12-09
Support PKCS#11 for mutual TLS on Unix platforms

graebm:
Add bindings for new PKCS#11 functionality.

Related: awslabs/aws-c-io#451

### DIFF
--- a/.builder/actions/aws_crt_python.py
+++ b/.builder/actions/aws_crt_python.py
@@ -4,6 +4,7 @@ import argparse
 import os.path
 import pathlib
 import subprocess
+import re
 import sys
 import tempfile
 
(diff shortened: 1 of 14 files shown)


.builder/actions/aws_crt_python.py
```diff
@@ -37,34 +38,183 @@ def run(self, env):
...
+
+        # create token
+        token_label = 'my-token'
+        pin = '0000'
+        init_token_result = self._exec_softhsm2_util('--init-token', '--free', '--label', token_label,
+                                                     '--pin', pin, '--so-pin', '0000')
+
+        # figure out which slot the token ended up in.
+        #
+        # older versions of SoftHSM2 (2.1.0) make us pass --slot number to the --import command.
```
sbSteveK, line 100: makes*

.builder/actions/aws_crt_python.py
```diff
@@ -37,34 +38,183 @@ def run(self, env):
...
+        # (newer version let us pass --label name instead)
```
sbSteveK, line 101: lets*

awscrt/io.py
```diff
@@ -276,7 +290,65 @@ def create_client_with_mtls(cert_buffer, key_buffer):
...
+
+    @staticmethod
+    def create_client_with_mtls_pkcs11(*,
+                                       pkcs11_lib: 'Pkcs11Lib',
+                                       user_pin: str,
+                                       slot_id: int = None,
+                                       token_label: str = None,
+                                       private_key_label: str = None,
+                                       cert_file_path: str = None,
+                                       cert_file_contents=None):
```
sbSteveK, line 303: cert_file_contents=None -> cert_file_contents = None
add spaces for uniformity with other arguments?

source/io.c
```diff
@@ -453,7 +523,8 @@ PyObject *aws_py_client_tls_ctx_new(PyObject *self, PyObject *args) {
             goto ctx_options_failure;
         }
     }
-    if (ca_buffer_len > 0) {
+
+    if (ca_buffer) {
```
sbSteveK, line 527: 460 uses 
if (certificate_buffer != NULL) 
as the check vs. if(ca_buffer) here.

source/io.c
```diff
@@ -426,20 +439,77 @@ PyObject *aws_py_client_tls_ctx_new(PyObject *self, PyObject *args) {
...
+            &pkcs11_cert_file_path_len,
+            &pkcs11_cert_file_contents,
+            &pkcs11_cert_file_contents_len)) {
         return NULL;
     }
 
     struct aws_tls_ctx_options ctx_options;
     AWS_ZERO_STRUCT(ctx_options);
-    if (certificate_buffer_len > 0 && private_key_buffer_len > 0) {
+    if (certificate_buffer != NULL) {
```
sbSteveK, line 460: 527 uses
if(ca_buffer)
as the check vs. if (certificate_buffer != NULL).

.builder/actions/aws_crt_python.py
graebm: tweaked sentence

sbSteveK (approved):

awscrt/io.py
graebm: the autoformatter made it look like this.

The reason they're inconsistent is that every argument **except** `cert_file_contents` has a [type hint](https://docs.python.org/3/library/typing.html). So the autoformatter is like "this is how I format default values with type hints, and this is how I format default values without type hints". It's too much effort to fight the autoformatter.

I didn't put a type hint on `cert_file_contents` because it just needs to be a "bytes-like object" and there's no appropriate type hint for that right now: see https://bugs.python.org/issue27501

source/io.c
graebm: yeah, our code is pretty inconsistent on explicit comparisons `if (x != NULL)` 
vs
implicit comparisons `if (x)`

I was just trying to match whatever code was nearest
but I'll make the function more consistent, and favor explicit comparisons

awscrt/io.py
```diff
@@ -301,7 +373,6 @@ def create_client_with_mtls_pkcs12(pkcs12_filepath, pkcs12_password):
         opt = TlsContextOptions()
         opt.pkcs12_filepath = pkcs12_filepath
         opt.pkcs12_password = pkcs12_password
-        opt.verify_peer = True
```
bretambrose, line 304: What motivated this?

bretambrose (approved):

awscrt/io.py
graebm: The default `TlsContextOptions()` constructor already sets this true
it was weird that this one factory function was setting it explicitly when the other factory functions weren't

PR #334, opened 2022-02-15
Connection builder simplification

TwistedTwigleg:
*Description of changes:*

These changes make it where it's not required to create and manage a DefaultHostResolver, EventLoopGroup, or ClientBootstrap to make a connection. Instead, default singletons have been added to handle this use case.

The constructors of several classes have also been altered to allow passing a ClientBootstrap to be optional rather than required. When optional, the ClientBootstrap singleton will be used.
The MQTT connection class will free the singletons if no ClientBootstrap is passed, making the samples not require creating or directly referencing the ClientBootstrap at all.

Finally, the singletons were added to the testing to ensure they work as expected moving forward.

_______

*By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.*

### DIFF
--- a/awscrt/auth.py
+++ b/awscrt/auth.py
@@ -125,7 +125,7 @@ class AwsCredentialsProvider(AwsCredentialsProviderBase):
         self._binding = binding
 
     @classmethod
-    def new_default_chain(cls, client_bootstrap):
+    def new_default_chain(cls, client_bootstrap=None):
         """
         Create the default provider chain used by most AWS SDKs.
 
@@ -137,12 +137,16 @@ class AwsCredentialsProvider(AwsCredentialsProviderBase):
         4.  (conditional, on by default) EC2 Instance Metadata
 
         Args:
-            client_bootstrap (ClientBootstrap): Client bootstrap to use when initiating socket connection.
+            client_bootstrap (Optional[ClientBootstrap]): Client bootstrap to use when initiating socket connection.
+                If not set, uses the default static ClientBootstrap instead.
 
         Returns:
             AwsCredentialsProvider:
         """
-        assert isinstance(client_bootstrap, ClientBootstrap)
+        assert isinstance(client_bootstrap, ClientBootstrap) or client_bootstrap is None
+
+        if client_bootstrap is None:
+            client_bootstrap = ClientBootstrap.get_or_create_static_default()
 
         binding = _awscrt.credentials_provider_new_chain_default(client_bootstrap)
         return cls(binding)
@@ -170,7 +174,7 @@ class AwsCredentialsProvider(AwsCredentialsProviderBase):
     @classmethod
     def new_profile(
             cls,
-            client_bootstrap,
+            client_bootstrap=None,
             profile_name=None,
             config_filepath=None,
             credentials_filepath=None):
@@ -179,7 +183,8 @@ class AwsCredentialsProvider(AwsCredentialsProviderBase):
         loaded from the aws credentials file.
 
         Args:
-            client_bootstrap (ClientBootstrap): Client bootstrap to use when initiating socket connection.
+            client_bootstrap (Optional[ClientBootstrap]): Client bootstrap to use when initiating socket connection.
+                If not set, uses the static default ClientBootstrap instead.
 
             profile_name (Optional[str]): Name of profile to use.
                 If not set, uses value from AWS_PROFILE environment variable.
@@ -196,10 +201,13 @@ class AwsCredentialsProvider(AwsCredentialsProviderBase):
         Returns:
             AwsCredentialsProvider:
         """
-        assert isinstance(client_bootstrap, ClientBootstrap)
+        assert isinstance(client_bootstrap, ClientBootstrap) or client_bootstrap is None
         assert isinstance(profile_name, str) or profile_name is None
         assert isinstance(config_filepath, str) or config_filepath is None
         assert isinstance(credentials_filepath, str) or credentials_filepath is None
+
+        if client_bootstrap is None:
+            client_bootstrap = ClientBootstrap.get_or_create_static_default()
 
         binding = _awscrt.credentials_provider_new_profile(
             client_bootstrap, profile_name, config_filepath, credentials_filepath)
--- a/awscrt/io.py
+++ b/awscrt/io.py
@@ -94,7 +94,6 @@ class EventLoopGroup(NativeResource):
         self.shutdown_event = shutdown_event
         self._binding = _awscrt.event_loop_group_new(num_threads, is_pinned, cpu_group, on_shutdown)
 
-
     @staticmethod
     def get_or_create_static_default():
         global _static_event_loop_group
@@ -103,23 +102,22 @@ class EventLoopGroup(NativeResource):
                 return _static_event_loop_group
 
         os_cpu_count = os.cpu_count()
-        if os_cpu_count == None:
+        if os_cpu_count is None:
             os_cpu_count = 1 # could not find the processor count - default to 1
         else:
-            #os.cpu_count returns all cores on the system (including virtual ones)
-            #so we divide by two to avoid using all cores on the entire system.
-            os_cpu_count = max(1, int(os_cpu_count/2))
+            # os.cpu_count returns all cores on the system (including virtual ones)
+            # so we divide by two to avoid using all cores on the entire system.
+            os_cpu_count = max(1, int(os_cpu_count / 2))
 
         _static_event_loop_group = EventLoopGroup(os_cpu_count)
         return _static_event_loop_group
-
 
     @staticmethod
     def release_static_default():
         global _static_event_loop_group
         if '_static_event_loop_group' in globals():
             if isinstance(_static_event_loop_group, EventLoopGroup):
-                del _static_event_loop_group
+                del _static_event_loop_group # lgtm [py/unnecessary-delete]
 
 
 class HostResolverBase(NativeResource):
@@ -142,7 +140,6 @@ class DefaultHostResolver(HostResolverBase):
         super().__init__()
         self._binding = _awscrt.host_resolver_new_default(max_hosts, event_loop_group)
 
-
     @staticmethod
     def get_or_create_static_default():
         global _static_default_host_resolver
@@ -152,12 +149,11 @@ class DefaultHostResolver(HostResolverBase):
         _static_default_host_resolver = DefaultHostResolver(EventLoopGroup.get_or_create_static_default())
         return _static_default_host_resolver
 
-
     @staticmethod
     def release_static_default():
         global _static_default_host_resolver
         if isinstance(_static_default_host_resolver, DefaultHostResolver):
-            del _static_default_host_resolver
+            del _static_default_host_resolver # lgtm [py/unnecessary-delete]
 
 
 class ClientBootstrap(NativeResource):
@@ -188,7 +184,6 @@ class ClientBootstrap(NativeResource):
         self.shutdown_event = shutdown_event
         self._binding = _awscrt.client_bootstrap_new(event_loop_group, host_resolver, on_shutdown)
 
-
     @staticmethod
     def get_or_create_static_default():
         global _static_client_bootstrap
@@ -200,12 +195,11 @@ class ClientBootstrap(NativeResource):
             DefaultHostResolver.get_or_create_static_default())
         return _static_client_bootstrap
 
-
     @staticmethod
     def release_static_default():
         global _static_client_bootstrap
         if isinstance(_static_client_bootstrap, ClientBootstrap):
-            del _static_client_bootstrap
+            del _static_client_bootstrap # lgtm [py/unnecessary-delete]
 
 
 def _read_binary_file(filepath):
--- a/awscrt/io.py
+++ b/awscrt/io.py
@@ -126,7 +126,7 @@ class DefaultHostResolver(HostResolverBase):
         with DefaultHostResolver._static_host_resolver_lock:
             if DefaultHostResolver._static_host_resolver is None:
                 DefaultHostResolver._static_host_resolver = DefaultHostResolver(
-                    EventLoopGroup.get_or_create_static_default)
+                    EventLoopGroup.get_or_create_static_default())
             return DefaultHostResolver._static_host_resolver
 
     @staticmethod
@@ -169,7 +169,7 @@ class ClientBootstrap(NativeResource):
     @staticmethod
     def get_or_create_static_default():
         with ClientBootstrap._static_client_bootstrap_lock:
-            if ClientBootstrap._static_client_bootstrap_lock is None:
+            if ClientBootstrap._static_client_bootstrap is None:
                 ClientBootstrap._static_client_bootstrap = ClientBootstrap(
                     EventLoopGroup.get_or_create_static_default(),
                     DefaultHostResolver.get_or_create_static_default())
--- a/awscrt/io.py
+++ b/awscrt/io.py
@@ -87,22 +87,15 @@ class EventLoopGroup(NativeResource):
 
     @staticmethod
     def get_or_create_static_default():
-        EventLoopGroup._static_event_loop_group_lock.acquire()
-        if EventLoopGroup._static_event_loop_group is not None:
-            if isinstance(EventLoopGroup._static_event_loop_group, EventLoopGroup):
-                EventLoopGroup._static_event_loop_group_lock.release()
-                return EventLoopGroup._static_event_loop_group
-        EventLoopGroup._static_event_loop_group = EventLoopGroup()
-        EventLoopGroup._static_event_loop_group_lock.release()
-        return EventLoopGroup._static_event_loop_group
+        with EventLoopGroup._static_event_loop_group_lock:
+            if EventLoopGroup._static_event_loop_group is None:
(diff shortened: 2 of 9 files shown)


awscrt/auth.py
```diff
@@ -196,11 +201,14 @@ def new_profile(
...
         """
-        assert isinstance(client_bootstrap, ClientBootstrap)
+        assert isinstance(client_bootstrap, ClientBootstrap) or client_bootstrap is None
         assert isinstance(profile_name, str) or profile_name is None
         assert isinstance(config_filepath, str) or config_filepath is None
         assert isinstance(credentials_filepath, str) or credentials_filepath is None
 
+        if client_bootstrap is None:
+            client_bootstrap = ClientBootstrap.get_or_create_static_default()
+
```
xiazhvera, line 211: Where would we release the static client bootstrap once it get created here? Currently I only find the release happens in the MQTT Connection, but I think we can also use AwsCredentialsProvider for other purpose.

awscrt/mqtt.py
```diff
@@ -435,6 +440,12 @@ def on_disconnect():
 
         try:
             _awscrt.mqtt_client_connection_disconnect(self._binding, on_disconnect)
+
+            if self.using_static_defaults:
+                ClientBootstrap.release_static_default()
+                DefaultHostResolver.release_static_default()
+                EventLoopGroup.release_static_default()
+
```
xiazhvera, line 448: Correct me if I missed anything, I think it might be better to have Client to handle the release like what we have in crt-java, since the bootstrap is referenced in Client class instead of the Connection.

awscrt/auth.py
TwistedTwigleg: I *think*, and it is entirely possible I am wrong here, that Python's garbage collector would free the singleton-like object once the module is unloaded, so it shouldn't need to be freed manually by the user.

That said though, I'm not entirely sure. I read a bunch about various singleton patterns in Python, but unfortunately the part about how memory is managed was rather unclear.

I probably should take another look at the design and confirm that it's being freed by unloading the module to ensure that it is indeed being garbage collected.

awscrt/mqtt.py
TwistedTwigleg: Good call! I'll adjust it 👍

awscrt/auth.py
TwistedTwigleg: I just checked and can confirm that when the module is unloaded and automatically garbage collected, it will free the singleton automatically. I added the following code to the end of the PubSub sample:

~~~
print ("Client bootstrap:")
print (io._static_client_bootstrap)

from importlib import reload
io = reload(io)
print ("Client bootstrap after reload:")
print (io._static_client_bootstrap)
~~~

And it reports that the object is cleared:

~~~
Client bootstrap:
<awscrt.io.ClientBootstrap object at 0x10a86c300>
Client bootstrap after reload:
None
~~~

xiazhvera (approved):

awscrt/io.py
```diff
@@ -83,6 +94,31 @@ def on_shutdown():
...
+            if isinstance(_static_event_loop_group, EventLoopGroup):
+                return _static_event_loop_group
+
+        os_cpu_count = os.cpu_count()
+        if os_cpu_count is None:
+            os_cpu_count = 1  # could not find the processor count - default to 1
+        else:
+            # os.cpu_count returns all cores on the system (including virtual ones)
+            # so we divide by two to avoid using all cores on the entire system.
+            os_cpu_count = max(1, int(os_cpu_count / 2))
```
graebm, line 110: Just create [`EventLoopGroup()`](https://github.com/awslabs/aws-crt-python/blob/cc651452fa1900c140fedde0be65962b5b897396/awscrt/io.py#L65) with no args to get default settings. Don't copy/paste the finicky default-picking logic here. If we modified the default-picking logic in C to be smarter, then this would override it

awscrt/io.py
```diff
@@ -83,6 +94,31 @@ def on_shutdown():
         self.shutdown_event = shutdown_event
         self._binding = _awscrt.event_loop_group_new(num_threads, is_pinned, cpu_group, on_shutdown)
 
+    @staticmethod
+    def get_or_create_static_default():
+        global _static_event_loop_group
+        if '_static_event_loop_group' in globals():
+            if isinstance(_static_event_loop_group, EventLoopGroup):
+                return _static_event_loop_group
```
graebm, line 102: gotta use locks when checking the singleton 

otherwise it's theoretically possible for two threads to call this at once, both threads see that no singleton exists, and both threads go ahead and create one

awscrt/io.py
```diff
@@ -83,6 +94,31 @@ def on_shutdown():
...
+
+        _static_event_loop_group = EventLoopGroup(os_cpu_count)
+        return _static_event_loop_group
+
+    @staticmethod
+    def release_static_default():
+        global _static_event_loop_group
+        if '_static_event_loop_group' in globals():
+            if isinstance(_static_event_loop_group, EventLoopGroup):
+                del _static_event_loop_group  # lgtm [py/unnecessary-delete]
```
graebm, line 120: just set the variable `= None`
you don't need to delete the existence of the variable

awscrt/auth.py
graebm: Yeah, Python is a garbage-collected language so you usually don't release or delete anything explicitly, just let the reference drop.

graebm: also, mild warning about `importlib.reload`. I wouldn't get into the habit of using that with aws-crt-python. The [docs](https://docs.python.org/3/library/importlib.html#importlib.reload) warn:

>  In many cases extension modules are not designed to be initialized more than once, and may fail in arbitrary ways when reloaded.

We haven't really tested reloading the C parts, and if something isn't tested it's probably broken. Years ago I wondered if reloading was something we had to support, but saw that that statement in the docs as a green light to ignore it

awscrt/io.py
```diff
@@ -39,6 +40,16 @@ def init_logging(log_level, file_name):
...
 
 
+"""
+Static Defaults (singletons - or as close as possible in Python)
+
+Will be garbage collected when program ends or module is unloaded
+"""
+_static_client_bootstrap = None
+_static_default_host_resolver = None
+_static_event_loop_group = None
```
graebm, line 50: trivial: one of these has the word "default" in it but the others don't. be consistent one way or the other

awscrt/mqtt.py
```diff
@@ -244,6 +244,9 @@ class Connection(NativeResource):
 
         proxy_options (Optional[awscrt.http.HttpProxyOptions]):
             Optional proxy options for all connections.
+
+        using_static_default (Optional[bool]):
+            If true, the connection will clean the static default ClientHandler, EventLoopGroup, and HostResolver.
```
graebm, line 249: this isn't needed anymore, remove

test/test_io.py
```diff
@@ -25,12 +25,41 @@ def test_shutdown_complete(self):
...
+        EventLoopGroup.release_static_default()
+        self.assertTrue(shutdown_event.wait(TIMEOUT))
+
 
 class DefaultHostResolverTest(NativeResourceTest):
     def test_init(self):
         event_loop_group = EventLoopGroup()
         host_resolver = DefaultHostResolver(event_loop_group)
 
+    def test_init_singleton(self):
```
graebm, line 51: Please add a test for MQTT where we let it use the singletons, to be sure it all works end to end in a complex case?

test/test_io.py
```diff
@@ -43,6 +72,27 @@ def test_create_destroy(self):
...
+        client_one = ClientBootstrap.get_or_create_static_default()
+        client_two = ClientBootstrap.get_or_create_static_default()
+        self.assertTrue(client_one == client_two)
+        ClientBootstrap.release_static_default()
+        EventLoopGroup.release_static_default()
+        DefaultHostResolver.release_static_default()
+
 
 class ClientTlsContextTest(NativeResourceTest):
     def test_init_defaults(self):
```
graebm, line 98: add support for static-defaults everywhere EventLoopGroup, HostResolverBase, and ClientBootstrap are taken:
- [http.HttpClientConnection.new()](https://github.com/awslabs/aws-crt-python/blob/cc651452fa1900c140fedde0be65962b5b897396/awscrt/http.py#L85)
    -  ugh, looks like once upon a time  **someone** already tried adding some hideous version of static defaults that would create NEW objects if the bootstrap wasn't specified. That is horrible. That is a bloated waste of resource. Oh, [that was me](https://github.com/awslabs/aws-crt-python/blame/cc651452fa1900c140fedde0be65962b5b897396/awscrt/http.py#L129) 😐
- [mqtt.Client()](https://github.com/awslabs/aws-crt-python/blob/cc651452fa1900c140fedde0be65962b5b897396/awscrt/mqtt.py#L141)
- [s3.S3Client()](https://github.com/awslabs/aws-crt-python/blob/cc651452fa1900c140fedde0be65962b5b897396/awscrt/s3.py#L90)
- [eventstream.rpc.ClientConnection.connect()](https://github.com/awslabs/aws-crt-python/blob/cc651452fa1900c140fedde0be65962b5b897396/awscrt/eventstream/rpc.py#L261)

awscrt/mqtt.py
```diff
@@ -13,7 +13,7 @@
 from awscrt import NativeResource
 import awscrt.exceptions
 from awscrt.http import HttpProxyOptions, HttpRequest
-from awscrt.io import ClientBootstrap, ClientTlsContext, SocketOptions
+from awscrt.io import ClientBootstrap, ClientTlsContext, SocketOptions, EventLoopGroup, DefaultHostResolver
```
graebm, line 16: [trivial] this change isn't needed anymore

awscrt/mqtt.py
```diff
@@ -266,7 +269,8 @@ def __init__(self,
                  use_websockets=False,
                  websocket_proxy_options=None,
                  websocket_handshake_transform=None,
-                 proxy_options=None
+                 proxy_options=None,
+                 using_static_defaults=None
```
graebm, line 273: this isn't needed anymore, remove

test/test_io.py
```diff
@@ -43,6 +72,27 @@ def test_create_destroy(self):
...
+
+        self.assertTrue(bootstrap_shutdown_event.wait(TIMEOUT))
+
+    def test_init_singleton_is_singleton(self):
+        client_one = ClientBootstrap.get_or_create_static_default()
+        client_two = ClientBootstrap.get_or_create_static_default()
+        self.assertTrue(client_one == client_two)
+        ClientBootstrap.release_static_default()
+        EventLoopGroup.release_static_default()
+        DefaultHostResolver.release_static_default()
```
graebm, line 94: Add these 3 `release_static_default()` calls to this function:
[awscrt._test.check_for_leaks()](https://github.com/awslabs/aws-crt-python/blob/cc651452fa1900c140fedde0be65962b5b897396/awscrt/_test.py#L88)

rambling background about cleanup in aws-crt-python...

aws-crt-python is different than the other language bindings in that it doesn't try too hard to clean things up on exit, so there's no pre-existing "ok clean it all up" function for the library. But we do have this private function we call at the end of every test that checks for memory leaks, so let's put it there for now.

Some background on why we don't even try to clean up...
We used to! But Python developers like to write simple applications that exit the moment they get what they want (or exit immediately because of error), and we kept getting crashes when the library tried to clean things up on exit but the Python application was still running stuff on another thread. So now we try and tolerate ["sloppy shutdown"](https://github.com/awslabs/aws-crt-python/pull/116)

TwistedTwigleg:
Thanks for the reviews! I adjusted the code accordingly 👍

awscrt/io.py
```diff
@@ -96,28 +87,22 @@ def on_shutdown():
...
-        else:
-            # os.cpu_count returns all cores on the system (including virtual ones)
-            # so we divide by two to avoid using all cores on the entire system.
-            os_cpu_count = max(1, int(os_cpu_count / 2))
-
-        _static_event_loop_group = EventLoopGroup(os_cpu_count)
-        return _static_event_loop_group
+        EventLoopGroup._static_event_loop_group_lock.acquire()
+        if EventLoopGroup._static_event_loop_group is not None:
+            if isinstance(EventLoopGroup._static_event_loop_group, EventLoopGroup):
```
graebm, line 92: [style] this insinstance() check isn't necessary. this is a private member, you can assume it's the correct type

in general, python programmers don't check that things are the expected type. They just go ahead and expect things to work, and if they don't work then an exception will bubble up

I know you've seen a lot of isinstance() checks in this codebase, but we only have those to check arguments that the user passes into public functions, and we do that because we pass a lot of things down to C, and the exceptions and stack traces aren't as clear to the user when they come from C, so we make sure the exception happens early

awscrt/io.py
```diff
@@ -96,28 +87,22 @@ def on_shutdown():
...
     @staticmethod
     def release_static_default():
-        global _static_event_loop_group
-        if '_static_event_loop_group' in globals():
-            if isinstance(_static_event_loop_group, EventLoopGroup):
-                del _static_event_loop_group  # lgtm [py/unnecessary-delete]
+        EventLoopGroup._static_event_loop_group_lock.acquire()
+        if EventLoopGroup._static_event_loop_group is not None:
+            if isinstance(EventLoopGroup._static_event_loop_group, EventLoopGroup):
+                EventLoopGroup._static_event_loop_group = None
```
graebm, line 104: [trivial] why bother checking if it's none, and what type it is? just set it `= None` no matter what

awscrt/io.py
```diff
@@ -96,28 +87,22 @@ def on_shutdown():
...
-        _static_event_loop_group = EventLoopGroup(os_cpu_count)
-        return _static_event_loop_group
+        EventLoopGroup._static_event_loop_group_lock.acquire()
+        if EventLoopGroup._static_event_loop_group is not None:
+            if isinstance(EventLoopGroup._static_event_loop_group, EventLoopGroup):
+                EventLoopGroup._static_event_loop_group_lock.release()
+                return EventLoopGroup._static_event_loop_group
+        EventLoopGroup._static_event_loop_group = EventLoopGroup()
+        EventLoopGroup._static_event_loop_group_lock.release()
+        return EventLoopGroup._static_event_loop_group
```
graebm, line 97: as an aside: this still isn't thread safe, though it won't matter any more once you use the with statement

it's theoretically possible that someone calls `EventLoopGroup.release_static_default()` from another thread in between the lock being released, and `EventLoopGroup._static_event_loop_group` being returned. In which case `None` would end up being returned.

The thread safe way to do it when manually releasing a lock would be like:
```
tmp = EventLoopGroup._static_event_loop_group
EventLoopGroup._static_event_loop_group_lock.release()
return tmp
``` 

so that even if the value of `EventLoopGroup._static_event_loop_group` changes them *moment* you release the lock, you still return what you'd meant to return

awscrt/mqtt.py
```diff
@@ -131,19 +131,22 @@ class Client(NativeResource):
 
     Args:
         bootstrap (ClientBootstrap): Client bootstrap to use when initiating new socket connections.
+            If None is provided, the default singleton is used.
```
graebm, line 134: same thing with `Optional[ClientBootstrap]`

test/test_io.py
```diff
@@ -78,20 +72,12 @@ def test_create_destroy_singleton(self):
...
         ClientBootstrap.release_static_default()
-        EventLoopGroup.release_static_default()
-        DefaultHostResolver.release_static_default()
-
         self.assertTrue(bootstrap_shutdown_event.wait(TIMEOUT))
 
     def test_init_singleton_is_singleton(self):
         client_one = ClientBootstrap.get_or_create_static_default()
         client_two = ClientBootstrap.get_or_create_static_default()
-        self.assertTrue(client_one == client_two)
```
graebm: bring back this assert(). you're not actually testing anything anymore :)

test/test_mqtt.py
```diff
@@ -213,5 +213,111 @@ def on_sub_message(topic, payload):
         connection.disconnect().result(TIMEOUT)
 
 
+class MqttConnectionSingletonTest(NativeResourceTest):
+    TEST_TOPIC = '/test/me/senpai'
+    TEST_MSG = 'NOTICE ME!'.encode('utf8')
+
+    def _create_connection(self, auth_type=AuthType.CERT_AND_KEY):
+        config = Config(auth_type)
```
graebm, line 221: [style] try to re-use more code.

instead of copy/pasting 200 lines of code, you could modify the `def _create_connection()` helper function with an additional arg. And just 1 test new actual test case like `test_connect_disconnect_with_static_singletons` is fine

```
def _create_connection(self, auth_type=AuthType.CERT_AND_KEY, use_static_singletons=False):
    bootstrap = None
    if not use_static_singletons:
        elg = EventLoopGroup()
        resolver = DefaultHostResolver(elg)
        bootstrap = ClientBootstrap(elg, resolver)
    ...        
```

awscrt/s3.py
```diff
@@ -55,6 +55,7 @@ class S3Client(NativeResource):
 
     Keyword Args:
         bootstrap (ClientBootstrap): Client bootstrap to use when initiating socket connection.
+            If None is provided, the default singleton is used.
```
graebm, line 58: same thing with `(Optional[ClientBootstrap])`

graebm (approved):
🚀

TwistedTwigleg:

PR #336, opened 2022-03-01
Adapt new input stream API

TingDaoK:
*Issue #, if available:*

*Description of changes:*


source/io.c
```diff
@@ -698,25 +698,25 @@ PyObject *aws_py_tls_connection_options_set_server_name(PyObject *self, PyObject
 /* aws_input_stream implementation for accessing Python I/O classes */
 struct aws_input_stream_py_impl {
     struct aws_input_stream base;
+    struct aws_allocator *allocator;
 
     bool is_end_of_stream;
 
     /* Weak reference proxy to python self. */
     PyObject *self_proxy;
```
graebm, line 706: "Weak reference proxy to python self"
😬

We got some real bad lifetime management going on here, due to the fact that none of this stuff used to be refcounted, and now's the time to unwind it.

Lifetime management currently works like this:
`python class awscrt.io.InputStream`
* keeps alive `python class IOStream`
* keeps alive `C struct aws_input_stream_py_impl`

`C struct aws_input_stream_py_impl`
* only has a **WEAK** reference to its own python class, assuming the python class can't go away before its `_binding`
* and accomplish this by painfully keeping the python class alive everywhere that uses it

but now that the C input stream can be kept alive longer than the python class, we're in trouble

We need to keep python stuff alive as long as the C part is alive, but can't make a circular reference.

I have ideas.... let's chat. too slow to write down

source/io.c
```diff
@@ -826,24 +852,18 @@ static struct aws_input_stream *aws_input_stream_new_from_py(PyObject *py_self)
...
     struct aws_input_stream_py_impl *impl = aws_mem_calloc(alloc, 1, sizeof(struct aws_input_stream_py_impl));
     if (!impl) {
         return NULL;
     }
 
-    impl->base.allocator = alloc;
+    impl->allocator = alloc;
     impl->base.vtable = &s_aws_input_stream_py_vtable;
-    impl->base.impl = impl;
-    impl->self_proxy = PyWeakref_NewProxy(py_self, NULL);
```
graebm, line 838: SHOULD we still do some kind of weakref here?
Is Python allowed to move the Python object in memory if we don't tell it that we're keeping C references via INCREF or weakref or anything?

(remind me to look this up after)

graebm: nevermind. what you're doing seems safe. The docs on C-extensions talk about "borrowing" a reference without calling Py_INCREF, seems like you just need to be very careful

https://docs.python.org/3/extending/extending.html#reference-counting-in-python

source/io.c
```diff
@@ -808,44 +803,48 @@ int s_aws_input_stream_py_get_status(struct aws_input_stream *stream, struct aws
...
     (void)out_length;
-    return AWS_ERROR_UNIMPLEMENTED;
+    return aws_raise_error(AWS_ERROR_UNIMPLEMENTED);
+}
+
+struct aws_input_stream *s_aws_input_stream_py_acquire(struct aws_input_stream *stream) {
+    if (stream) {
+        /*************** GIL ACQUIRE ***************/
+        PyGILState_STATE state;
+        if (aws_py_gilstate_ensure(&state)) {
```
graebm, line 813: optional optimization: instead of acquiring the GIL every time C refcount changes, we could only acquire the GIL and do Py_INCREF when refcount goes from 0->1, and vice-versa when C refcount goes from 1->0

source/io.c
```diff
@@ -867,15 +867,20 @@ PyObject *aws_py_input_stream_new(PyObject *self, PyObject *args) {
         return NULL;
     }
 
-    struct aws_input_stream *stream = aws_input_stream_new_from_py(py_self);
-    if (!stream) {
-        return PyErr_AwsLastError();
+    if (!py_self || (py_self == Py_None)) {
+        aws_raise_error(AWS_ERROR_INVALID_ARGUMENT);
+        return NULL;
```
graebm, line 872: this isn't doing proper error reporting. previously, this was in a helper function that did aws_last_error() style reporting. But this is function is called directly from python and needs to raise a python exception

do something like:

```
PyErr_SetString(PyExc_TypeError, "InputStream cannot be None");
return NULL;
```

source/s3_meta_request.c
```diff
@@ -236,9 +236,13 @@ static void s_s3_request_on_finish(
 
     if (request_binding->input_body) {
         /* close the input file stream now, request has finished, we will not read from there anymore */
-        aws_input_stream_destroy(request_binding->input_body);
+        aws_input_stream_release(request_binding->input_body);
```
graebm, line 239: we don't need the `s3_meta_request_binding.input_body` variable at all anymore

delete it

when we create the input_body, we add it to the request. Then we can just release it, the request will keep it alive

source/s3_meta_request.c
```diff
@@ -236,9 +236,13 @@ static void s_s3_request_on_finish(
...
         request_binding->input_body = NULL;
     }
+    if (request_binding->copied_message) {
+        aws_http_message_release(request_binding->copied_message);
+        request_binding->copied_message = NULL;
+    }
```
graebm, line 245: two trivial things:

1) was this a memory leak you caught? I don't see similar code in the before-view
2) these 4 lines can just be `request_binding->copied_message = aws_http_message_release(request_binding_copied_message);`

source/s3_meta_request.c
```diff
@@ -311,14 +315,20 @@ static void s_s3_request_on_shutdown(void *user_data) {
     if (aws_py_gilstate_ensure(&state)) {
         return; /* Python has shut down. Nothing matters anymore, but don't crash */
     }
+    /* Clean thing up before invoking the callback, as the callback means everything has been shutdown, which is not
+     * true if we clean the resource after the callback */
```
graebm, line 319: nice comment 👍

source/s3_meta_request.c
TingDaoK: yes, it was uncovered by the input stream change

awscrt/http.py
```diff
@@ -330,13 +330,14 @@ def headers(self):
...
-        """InputStream: Binary stream of outgoing body."""
-        return _awscrt.http_message_get_body_stream(self._binding)
+        return self._body_stream
 
     @body_stream.setter
     def body_stream(self, stream):
+        wrapper = InputStream.wrap(stream)
+        _awscrt.http_message_set_body_stream(self._binding, wrapper)
+        self._body_stream = stream
         stream = InputStream.wrap(stream)
```
graebm, line 340: this can be simpler:
```
self._body_stream = InputStream.wrap(stream)
_awscrt.http_message_set_body_stream(self._binding, self._body_stream)
```

source/s3_meta_request.c
```diff
@@ -327,12 +320,14 @@ static void s_s3_request_on_shutdown(void *user_data) {
  * file-based python input stream for reporting the progress
  */
 struct aws_input_py_stream_file_impl {
+    struct aws_input_stream base;
+    struct aws_allocator *allocator;
```
graebm, line 324: don't need to store the allocator. in aws-crt-python there's only one global allocator

graebm (approved):
fix and (optimize and) ship

PR #337, opened 2022-03-03
check the error message to only handle the expected exception

TingDaoK:
fix Misleading error message when mqtt on_message_received callback raises a TypeError by checking the error message.


awscrt/mqtt.py
```diff
@@ -499,11 +499,14 @@ def subscribe(self, topic, qos, callback=None):
             def callback_wrapper(topic, payload, dup, qos, retain):
                 try:
                     callback(topic=topic, payload=payload, dup=dup, qos=QoS(qos), retain=retain)
-                except TypeError:
-                    # This callback used to have fewer args.
-                    # Try again, passing only those those args, to cover case where
-                    # user function failed to take forward-compatibility **kwargs.
-                    callback(topic=topic, payload=payload)
+                except TypeError as e:
+                    if "got an unexpected keyword argument" in str(e) and callback.__name__ in str(e):
```
ilevyor, line 503: Is there a more robust way to handle this? What if someone changes the string value of the error? I don't think they would expect that to affect people's code execution beyond logging.

awscrt/mqtt.py
```diff
@@ -499,11 +499,14 @@ def subscribe(self, topic, qos, callback=None):
...
+                        # This callback used to have fewer args.
+                        # Try again, passing only those those args, to cover case where
```
ilevyor, line 505:
```suggestion
                        # Try again, passing only those args, to cover case where
```

awscrt/mqtt.py
```diff
@@ -557,11 +560,15 @@ def on_message(self, callback):
...
```
ilevyor, line 566:
```suggestion
                        # Try again, passing only those args, to cover case where
```

awscrt/mqtt.py
```diff
@@ -322,6 +322,19 @@ def __init__(self,
...
+    def _check_try_backward_message_callback(self, callback):
+        # The callback used to have fewer args. Passing only those args, if it
+        # only has two args and no forward-compatibility to cover case where
+        # user function failed to take forward-compatibility **kwargs.
+
+        try_backward_callback = False
+        callback_sig = signature(callback)
+        try:
+            callback_sig.bind(topic='topic', payload='payload', dup=True, qos=QoS(1), retain=True)
+        except Exception as e:
```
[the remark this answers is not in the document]
graebm: optional: before this PR we didn't know it was possible to check signatures at all

now that we do, let's proactively check the callback will work with the old signature as well. maybe like:
```
try:
    # try new signature
    callback_sig.bind(topic='topic', payload='payload', dup=True, qos=QoS(1), retain=True)
    return False
except TypeError:
    # try old signature
    callback_sig.bind(topic='topic', payload='payload')
    return True
```

graebm (approved):
fix & ship

PR #356, opened 2022-04-27
Support Universal macOS Binary.

graebm:
**Issue #** https://github.com/aws/aws-iot-device-sdk-python-v2/issues/286

Our `universal2` wheels are currently broken. They don't have everything they need to run on both `x86_64` and `arm64`. We only compiled our dependencies for 1 architecture. The linker was warning us about this but we never noticed.

**Description of Changes:**
When building `universal2` wheels, compile dependencies for both `x86_64` and `arm64`, then use `lipo` to glue them together into a universal binary.

Also:
- Turn on linker warnings. This could have helped us catch these problems earlier.
- Use more specific build tmp folder to prevents collisions when building for multiple python versions on the same machine.
- Properly set `MACOSX_DEPLOYMENT_TARGET`

**Alternatives Explored:**
Experimented with compiling both architectures at the same time via `-DCMAKE_OSX_ARCHITECTURES=x86_64;arm64`, but there were big disadvantages to this way. Mainly, we can't pass any compile flag unless it's supported by both architectures. So we couldn't enable fancy stuff like SSE/AVX instructions for the x86_64 build. Seems better to compile for each architecture using the best possible flags, then just glue them together afterwards.

### DIFF
--- a/continuous-delivery/build-wheels-osx.sh
+++ b/continuous-delivery/build-wheels-osx.sh
@@ -5,7 +5,6 @@ set -e
 
 /Library/Frameworks/Python.framework/Versions/3.9/bin/python3 ./continuous-delivery/update-version.py
 
-export MACOSX_DEPLOYMENT_TARGET="10.9"
 /Library/Frameworks/Python.framework/Versions/3.6/bin/python3 setup.py bdist_wheel
 /Library/Frameworks/Python.framework/Versions/3.7/bin/python3 setup.py sdist bdist_wheel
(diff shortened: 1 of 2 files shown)


ilevyor (approved):

graebm:
Tested by building wheel on Intel Mac, then installing it on an M1 Mac and running the unit tests. Every test that wasn't skipped passed. The reason many were skipped is that I didn't want to put team credentials on my personal machine.

PR #376, opened 2022-07-22
Fix FreeBSD build

gearnode:
This pull request requires:
1. Merge the `aws-c-common` [pull request](https://github.com/awslabs/aws-c-common/pull/921).  -> done
2. Update the `aws-c-common` git module -> done

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -320,7 +320,10 @@ def awscrt_ext():
 
     if distutils.ccompiler.get_default_compiler() != 'msvc':
         extra_compile_args += ['-Wextra', '-Werror', '-Wno-strict-aliasing', '-std=gnu99']
-        extra_link_args += ['-Wl,-fatal_warnings']
+        if sys.platform.startswith('freebsd'):
+            extra_link_args += ['-Wl,-fatal-warnings']
+        else:
+            extra_link_args += ['-Wl,-fatal_warnings']
 
     return setuptools.Extension(
         '_awscrt',
--- a/setup.py
+++ b/setup.py
@@ -321,7 +321,7 @@ def awscrt_ext():
     if distutils.ccompiler.get_default_compiler() != 'msvc':
         extra_compile_args += ['-Wextra', '-Werror', '-Wno-strict-aliasing', '-std=gnu99']
         if sys.platform.startswith('freebsd'):
-            extra_link_args += ['-Wl,-fatal-warnings']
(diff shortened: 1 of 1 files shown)


gearnode:
The patch to build aws-c-common is now available on the main branch. I've updated the submodule in this repository. The pull request can be reviewed and merged.

gearnode:
ping @TingDaoK

setup.py
```diff
@@ -316,7 +316,7 @@ def awscrt_ext():
     if distutils.ccompiler.get_default_compiler() != 'msvc':
         extra_compile_args += ['-Wextra', '-Werror', '-Wno-strict-aliasing', '-std=gnu99']
         if sys.platform.startswith('freebsd'):
-            extra_link_args += ['-Wl,-fatal-warnings']
+            extra_link_args += ['-Wl,-fatal-warnings', '-lexecinfo']
         else:
             extra_link_args += ['-Wl,-fatal_warnings']
```
hxw, line 321: I think this ought to be `-falal-warnings` also or this will fail when using Clang on Linux.
I looked at both Clang and GNU linkers and neither showed using underscore in this option

hxw:
I think this patch may be all that is needed for me to complete my port of awscli2 to FreeBSD at:  https://reviews.freebsd.org/D35506

gearnode:
> I think this patch may be all that is needed for me to complete my port of awscli2 to FreeBSD at: https://reviews.freebsd.org/D35506

Otis and me working on the AWS CLI refactor on the FreeBSD port. Maybe we should join our efforts.

hxw:
> > I think this patch may be all that is needed for me to complete my port of awscli2 to FreeBSD at: https://reviews.freebsd.org/D35506
> 
> Otis and me working on the AWS CLI refactor on the FreeBSD port. Maybe we should join our efforts.
> 
> [1] https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=265360 [2] https://reviews.freebsd.org/D36519

The only real problem with my port of awscli2 is that make test fails because the test process does not use FreeBSD port of py-awscrt (which patches the linker problem) but downloads wheel files directly so get the linker errors. I am using awscli2 in daily use and have not had any problems with it so far. The current review updates py-awscrt to 0.14.6 which I have not tested yet

TingDaoK:
Fixed by https://github.com/awslabs/aws-crt-python/pull/473

PR #378, opened 2022-07-29
Developer Guide - Lifetime Management

graebm:
PROTIP for reviewers:

click `...` `->` `View file` on the `.md` file to see it as formatted HTML, with the images at the right spots.

or see click here to see it in the branch:
https://github.com/awslabs/aws-crt-python/blob/devguide3/docs/dev/writing-code.md

Please give feedback on:
- Am I being to verbose?
- Point out anything confusing. This guide is for YOU


docs/dev/writing-code.md
```diff
@@ -192,18 +201,195 @@ TODO: document when to use future vs callback
...
+    }
+    ```
+*   IF nothing else has a strong reference to `struct aws_event_loop_group`:
+    *   then it begins its shutdown process, and its memory is cleaned up when shutdown completes.
+*   ELSE something else has a strong reference to `struct aws_event_loop_group`:
+    *   so it won't begin its shutdown until the last reference is released.
+
+Note: In the past, the aws-c libs didn't have any reference counting for its C structs.
+You will still find older code in our Python bindings that tries to keep the entire dependency trees
+of Python objects alive via `Py_INCREF(x)` (TODO: remove needless complexity).
```
TwistedTwigleg, line 324: Minor nitpick: I'd place the TODO on a separate line on a bullet:

~~~
Note: In the past, the aws-c libs didn't have any reference counting for its C structs.
You will still find older code in our Python bindings that tries to keep the entire dependency trees
of Python objects alive via Py_INCREF(x). You can't always look at existing code to see "the right way"
of doing things.
  * (TODO: Remove this needless complexity in CRT)
~~~

docs/dev/writing-code.md
```diff
@@ -192,18 +201,195 @@ TODO: document when to use future vs callback
...
+
+You might be tempted to give `event_loop_group_binding` a strong reference to the `EventLoopGroup` instance.
+Then C could simply call a private member function like `EventLoopGroup._on_shutdown_complete()`.
+But this design creates a reference cycle (see image below):
+
+![Diagram of bad binding with reference cycle](./binding-bad-cycle.svg)
+
+### Option 1 - Pass Callbacks to C
+
+Most of our bindings work like this:
```
TwistedTwigleg, line 350: Question: Should this diagram have an options/builder that points to both the EventLoopGroup and the Callable? From my current understanding, generally a builder/settings class or similar is how we set the callable and create the EventLoopGroup.

docs/dev/writing-code.md
```diff
@@ -192,18 +201,195 @@ TODO: document when to use future vs callback
...
+
+![Diagram with private Core class](./binding-private-core.svg)
 
 # Writing C Code
 
-TODO: Write me.
+TODO:
+Talk about how our tests can and cannot check for leaks.
+Talk about which classes require a `close()` function, and which don't.
 Suggest writing as little C code as possible.
```
TwistedTwigleg, line 391: Super minor nitpick: I would reword this to be something like "as little Python-specific C code as possible", since I think that may better encapsulate the idea (since we generally write most big features in C for cross SDK/CRT use)

docs/dev/writing-code.md
```diff
@@ -192,18 +201,195 @@ TODO: document when to use future vs callback
...
+    *   The C callback finally deletes the `event_loop_group_binding`.
+        This struct only existed to keep two strong references,
+        but now they've both been released.
+
+### Option 2 - Private Core Class
+
+Another option is to build a private `_Core` class that contains anything
+that may need to outlive the main Python object.
+This technique hasn't actually been used, but the author of this doc
+thinks it might be a graceful way to build in the future:
```
TwistedTwigleg, line 382: Question: In this case, would there be two objects/classes that are managed by a bigger class that contains references to both? I'm just curious, as it sounds interesting, but I'm not totally sure I understand how it would work based on the picture...

TwistedTwigleg (commented):
Looks good! There was a few really minor things I commented on, but nothing major at all and everything looks good 👍 

Thanks for putting this together!

docs/dev/writing-code.md
```diff
@@ -192,18 +201,195 @@ TODO: document when to use future vs callback
...
+    *   Header file: `crt/aws-c-io/include/aws/io/event_loop.h`
+*   `event_loop_group_binding` - The "bindings" struct.
+    Holds a strong reference to the underlying native implementation (usually in a member variable named "native")
+    *   Lives in Python/C extension module: `_awscrt`
+    *   Source file: `source/io.c`
+*   `PyCapsule` - The Python object which "owns" the pointer to `event_loop_group_binding`.
+*   `EventLoopGroup` - The Python class that users create and interact with.
+    Holds a reference to the `PyCapsule` in a member variable named "_binding".
+    *   Lives in Python module: `awscrt.io`
+    *   Source File: `awscrt/io.py`
```
waahm7, line 261: Suggestion: Maybe we can utilize hyperlinks to make it easier to jump into code. However, I am not sure as aws & awslabs will have different links.

waahm7 (commented):
Looks good to me. It's not too verbose for me. 

Thanks for putting this together!

docs/dev/writing-code.md
```diff
@@ -192,18 +201,195 @@ TODO: document when to use future vs callback
...
+![Diagram of Simple Binding](./binding-simple.svg)
+
+Description of parts:
+*   `aws_event_loop_group` - The underlying native implementation struct,
+    which knows nothing about Python.
+    *   Lives in C library: `aws-c-io`
+    *   Header file: `crt/aws-c-io/include/aws/io/event_loop.h`
+*   `event_loop_group_binding` - The "bindings" struct.
+    Holds a strong reference to the underlying native implementation (usually in a member variable named "native")
+    *   Lives in Python/C extension module: `_awscrt`
```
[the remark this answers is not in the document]
sbSteveK: This is probably because when calling things in the crt, you use the '_awscrt' module.
e.g. to call subscribe from python I use '_awscrt.mqtt5_client_subscribe()' and not 'aws-crt-python.mqtt5_client_subscribe()'

docs/dev/writing-code.md
```diff
@@ -192,18 +201,195 @@ TODO: document when to use future vs callback
...
+and release the reference by calling `Py_DECREF(x)`.
+Note that EVERYTHING in Python is an object: even functions, even numbers, even `None`.
+
+In the aws-c libraries, you create a strong reference to a struct by calling
+its `_acquire(x)` function, and release by calling its `_release(x)` function.
+Within the aws-c libraries, structs keep each other alive as long as necessary using these functions.
+Not every struct in the aws-c libraries has `_acquire(x)` and `_release(x)` functions,
+only heap-allocated structs with complex or unpredictable lifetimes.
+Every struct bound to a Python class is considered to have an unpredictable lifetime.
+
```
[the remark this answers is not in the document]
graebm: added something along those lines...

docs/dev/writing-code.md
```diff
@@ -192,18 +201,195 @@ TODO: document when to use future vs callback
...
+C code can create a strong reference to a Python object by calling `Py_INCREF(x)`,
+and release the reference by calling `Py_DECREF(x)`.
+Note that EVERYTHING in Python is an object: even functions, even numbers, even `None`.
+
+In the aws-c libraries, you create a strong reference to a struct by calling
+its `_acquire(x)` function, and release by calling its `_release(x)` function.
+Within the aws-c libraries, structs keep each other alive as long as necessary using these functions.
+Not every struct in the aws-c libraries has `_acquire(x)` and `_release(x)` functions,
+only heap-allocated structs with complex or unpredictable lifetimes.
+Every struct bound to a Python class is considered to have an unpredictable lifetime.
```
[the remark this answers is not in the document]
graebm: I added a bunch more detail, then realized this is just the "terminology" section.

So I simplified this terminology section

and moved the extra details I'd written down into the "Writing C Code" section, which I hadn't really started on before...

docs/dev/writing-code.md
graebm: Yeah, there's a ton more complexity that goes into hooking up real callbacks.

I'm just trying to illustrate that we need to avoid circular references, and keep out any details I can which would distract from that.

docs/dev/writing-code.md
graebm: this is my TODO list man

what I meant here was: write Python instead of C wherever possible because it's like 20 lines of code to read a member variable from C

I went and added a new section earlier in the document about "Don't Add Features"

docs/dev/writing-code.md
graebm: elaborated

docs/dev/writing-code.md
graebm: I tried and it looked worse. I want this `Note` to be once block of text

docs/dev/writing-code.md
```diff
@@ -192,18 +201,195 @@ TODO: document when to use future vs callback
...
+Description of parts:
+*   `aws_event_loop_group` - The underlying native implementation struct,
+    which knows nothing about Python.
+    *   Lives in C library: `aws-c-io`
+    *   Header file: `crt/aws-c-io/include/aws/io/event_loop.h`
+*   `event_loop_group_binding` - The "bindings" struct.
+    Holds a strong reference to the underlying native implementation (usually in a member variable named "native")
+    *   Lives in Python/C extension module: `_awscrt`
+    *   Source file: `source/io.c`
+*   `PyCapsule` - The Python object which "owns" the pointer to `event_loop_group_binding`.
```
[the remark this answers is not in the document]
graebm: added (left to right) comment

docs/dev/writing-code.md
```diff
@@ -190,20 +199,270 @@ Don't use this pattern unless you're adding to a class where it's already in use
...
+that can detect and clean up reference cycles among normal Python objects.
+HOWEVER, any cycle involving a `Py_INCREF(x)` from C creates an undetectable cycle.
+You MUST NOT create reference cycles when designing bindings.
+
+### Capsule
+
+[PyCapsule](https://docs.python.org/3/extending/extending.html#using-capsules)
+lets us bind the lifetime of a C struct to the lifetime of a Python object.
+It's a Python object that holds a C pointer and a "destructor" function pointer.
+When Python cleans up the `PyCapsule`, the destructor function will be called.
```
sbSteveK, line 243: Could be useful to state that PyCapsule_New() creates a reference and if the object is returned to Python, once the object goes out of scope out there it will lose the reference causing it to call its destructor unless it's been INCREF'ed somewhere else. Or maybe that's super obvious so doesn't need mentioning?

docs/dev/writing-code.md
```diff
@@ -199,27 +199,31 @@ Don't use this pattern unless you're adding to a class where it's already in use
...
 
+## Don't Add Features
+
+When binding an API from the aws-c libraries, don't start adding extra features.
+If you're tempted to add any "special logic" that would be valuable to the other aws-crt language bindings,
+add that logic in the underlying aws-c library so that every other language can benefit.
+
+Even for trivial things like picking nice default values, put it in the underlying aws-c library.
+(see [Use `None` for Optional Arguments](#use-none-as-the-default-value-for-optional-arguments)).
+
```
TwistedTwigleg, line 210: Good call out 👍

TwistedTwigleg (approved):
Looks good to me 👍

docs/dev/writing-code.md
```diff
@@ -190,20 +199,270 @@ Don't use this pattern unless you're adding to a class where it's already in use
...
+            # define callable local function
+            def shutdown_callback():
+                ...do stuff...
+
+            self._binding = _awscrt.event_loop_group_new(shutdown_callback, ...)
+```
+
+When the final shutdown callback happens in C, the Python callable
+is invoked, and then the reference is released via `Py_DECREF(x)`.
```
sbSteveK, line 374: Might want to indicate that the callable that's passed to native gets Py_INCREF(x) to clarify keeping a strong reference there since we're pointing out the Py_DECREF(x) here.

sbSteveK (approved):

PR #397, opened 2022-09-22
Cognito credentials provider binding

bretambrose:
* Cognito credentials provider binding


awscrt/auth.py
```diff
@@ -306,6 +306,63 @@ def new_delegate(cls, get_credentials):
...
+
+        Args:
+            endpoint (str): Cognito Identity service regional endpoint to source credentials from.
+
+            identity (str): Cognito identity to fetch credentials relative to.
+
+            tls_ctx (ClientTlsContext): Client TLS context to use when querying cognito credentials by HTTP.
+
+            logins (Optional[List[Tuple[str, str]]]): List of tuples specifying pairs of identity provider name and
+                token values, representing established login contexts for identity authentication purposes
```
graebm, line 330: trivial: ~List~ -> `Sequence`

Some people like to use tuples instead of lists

awscrt/auth.py
```diff
@@ -306,6 +306,63 @@ def new_delegate(cls, get_credentials):
...
+        Returns:
+            AwsCredentialsProvider:
+        """
+
+        assert isinstance(endpoint, str)
+        assert isinstance(identity, str)
+        assert isinstance(tls_ctx, ClientTlsContext)
+        assert isinstance(logins, list) or logins is None
+        assert isinstance(custom_role_arn, str) or custom_role_arn is None
+        assert isinstance(http_proxy_options, awscrt.http.HttpProxyOptions) or http_proxy_options is None
```
graebm, line 350: trivial/debatable: this isn't really necessary anymore. We used to do this a lot because our error reporting from C was really bad, so catching errors earlier in Python was preferable. But our error-reporting from C has improved a lot.

and Python isn't a strictly typed language, and stuff like "duck typing" is not uncommon

and stuff like this makes it hard to accept a list or tuple

source/auth_credentials.c
```diff
@@ -665,3 +667,143 @@ PyObject *aws_py_credentials_provider_new_delegate(PyObject *self, PyObject *arg
...
+                AWS_ZERO_STRUCT(*login_entry);
+
+                if (!PyArg_ParseTuple(
+                        login_tuple_py,
+                        "s#s#",
+                        &login_entry->identity_provider_name.ptr,
+                        &login_entry->identity_provider_name.len,
+                        &login_entry->identity_provider_token.ptr,
+                        &login_entry->identity_provider_token.len)) {
+                    goto done;
```
graebm, line 748: trivial: You could set a more specific error by doing something like:
`PyErr_SetString(PyExc_TypeError, "'logins' has bad entry")`
or
`PyErr_Format(PyExc_TypeError, "logins[%zu] is invalid, should be type (str, str)");`

test/test_auth.py
```diff
@@ -220,6 +221,65 @@ def delegate_get_credentials():
...
+class CognitoCredentialsProviderProxyTest(NativeResourceTest):
+
+    def test_unauthenticated(self):
+        identity = os.environ.get('AWS_TESTING_COGNITO_IDENTITY')
+        tls_opts = awscrt.io.TlsContextOptions()
+        tls_context = awscrt.io.ClientTlsContext(tls_opts)
+        http_proxy_options = awscrt.http.HttpProxyOptions(
+            os.environ.get('AWS_TEST_HTTP_PROXY_HOST'),
+            int(os.environ.get('AWS_TEST_HTTP_PROXY_PORT', '0')))
+
```
graebm, line 270: trivial: the `skipUnless()` says PROXY_PORT is required, but then down here you have it defaulting to "0"

awscrt/auth.py
```diff
@@ -306,6 +306,63 @@ def new_delegate(cls, get_credentials):
         binding = _awscrt.credentials_provider_new_delegate(get_credentials)
         return cls(binding)
 
+    @classmethod
+    def new_cognito(
+            cls,
+            endpoint,
```
graebm, line 312: trivial/debatable: I'd suggest using keyword-only arguments like
```
def new_cognito(
    cls,
    *,
    endpoint,
    etc..
```

we couldn't use this feature back when we supported python 2 😡

and hey look!
it's [recommended in our developer guide](https://github.com/awslabs/aws-crt-python/blob/main/guides/dev/writing-code.md#functions-with-lots-of-arguments)!

graebm (commented):
nice

graebm (approved):
fix and ship

awscrt/auth.py
bretambrose: First reaction is to just drop the logins assert.  Does that seem reasonable?

test/test_auth.py
bretambrose: Well that's just so that it has a valid numerical value if someone does something dumb like set it to 'A'  The test will fail in that case which seems ok?

PR #404, opened 2022-10-27
Fix release pipeline

graebm:
**Issue:**
A docker image used in the release process started having git errors:
> fatal: detected dubious ownership in repository at '/aws-crt-python'

**Investigation:**
Git 2.35.2+ introduced additional [safety checks](https://github.blog/2022-04-12-git-security-vulnerability-announced/), that raise errors if the `.git/` folder isn't owned by the user running git. The release process runs this docker container while mounting the `aws-crt-python/` directory along with its `.git` folder. Within docker, the user is `root` by default, and when it saw the `.git/` folder owned by someone else, it freaked out.

We haven't seen this error in previous releases because we were accidentally using a very old build of that docker container. When we got a newer build, it had a newer version of git, which raised this error.

**Description of changes:**
- Fix error by running as current user within docker, instead of running as root.
- Unrelated: change "dev" version from `1.0.0-dev` -> `1.0.0.dev0`.  Python didn't like the old version string and normalized it to `1.0.0.dev0` anyway. This gets rid of some warnings at build time.
- Unrelated: remove some lines where we uselessly query version

**Other fixes considered:**
- Considered removing all git calls from the release pipeline, but this would have been a big change.
- Considered modifying all our docker images so they disabled these new safety checks via `git config --global --add safe.directory '*'`, but I wasn't comfortable lowering security like that given my shallow understanding of these issues.

### DIFF
--- a/continuous-delivery/build-wheels-manylinux2014-x86_64-jenkins.sh
+++ b/continuous-delivery/build-wheels-manylinux2014-x86_64-jenkins.sh
@@ -8,8 +8,11 @@ $(aws --region us-east-1 ecr get-login --no-include-email)
 
 docker pull $DOCKER_IMAGE
 
+# NOTE: run as current user to avoid git "dubious ownership" error,
+# and so that output artifacts don't belong to "root"
 docker run --rm \
     --mount type=bind,source=`pwd`,target=/aws-crt-python \
+    --user "$(id -u):$(id -g)" \
     --workdir /aws-crt-python \
     --entrypoint /bin/bash \
     $DOCKER_IMAGE \
--- a/awscrt/__init__.py
+++ b/awscrt/__init__.py
@@ -12,7 +12,7 @@ __all__ = [
     's3'
 ]
 
-__version__ = '1.0.0-dev'
+__version__ = '1.0.0.dev0'
 
 
 class NativeResource:
--- a/codebuild/cd/publish_to_prod_pypi.yml
+++ b/codebuild/cd/publish_to_prod_pypi.yml
@@ -19,7 +19,6 @@ phases:
       - cp -rv $CODEBUILD_SRC_DIR_aws_crt_python_osx/* dist/
       - ls -la dist/
       - cd aws-crt-python
-      - CURRENT_TAG_VERSION=$(git describe --tags | cut -f2 -dv)
       - python3 continuous-delivery/pull-pypirc.py prod
       - python3 -m twine upload -r pypi ../dist/*
   post_build:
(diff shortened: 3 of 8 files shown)


continuous-delivery/build-wheels-manylinux2014-x86_64-jenkins.sh
```diff
@@ -8,8 +8,11 @@ $(aws --region us-east-1 ecr get-login --no-include-email)
 
 docker pull $DOCKER_IMAGE
 
+# NOTE: run as current user to avoid git "dubious ownership" error,
+# and so that output artifacts don't belong to "root"
 docker run --rm \
     --mount type=bind,source=`pwd`,target=/aws-crt-python \
+    --user "$(id -u):$(id -g)" \
```
TingDaoK, line 15: I don't get why using current user will work. 

Isn't that because one jenkins job polls the repo and upload to S3? And the build runs on other jenkins job?

Are those two jekins job running on the same EC2, so it's the same user? Or they have the same user name and group name and git only checks for string match?

waahm7 (approved):

sbSteveK (approved):

continuous-delivery/build-wheels-manylinux2014-x86_64-jenkins.sh
xiazhvera: I think the problem here was that the we build within a docker container on the Jenkins. The docker was under root privilege by default so it failed to access the git source owns by the user here.

graebm: Turns out, when we unzip the file from S3, ownership of those files/folder is properly reassigned to the user. Git has no issue runnin on the unzipped files. The issue comes up when we MOUNT that unzipped folder into a docker container.

By default, inside a docker container, things run as `root`

But file ownership within that mounted folder is bananas. See:
```
[root@a8e45e5d0076 aws-crt-python]# ls -al
drwxr-xr-x 14 1778205 users  4096 Oct 26 15:56 .
drwxr-xr-x  1 root    root   4096 Oct 27 18:32 ..
drwxr-xr-x  9 1778205 users  4096 Oct 27 18:25 .git
drwxr-xr-x  3 1778205 users  4096 Oct 26 15:56 awscrt
...
```

Note how the parent directory is owned by `root`, but everything inside the mounted directory is owned by the user-id/group-id from host (the docker container doesn't know how to map the numeric user-id-num to a username, so it's just showing the number)

With git's new rules, when the `root` user runs git and see that the `.git/` folder is owned by someone else, it fails

I should also note that this behavior is inconsistent. When mounting a folder from my Mac, the docker container showed everything as being owned by `root` and git had no issues. The issue only occurred when mounting a folder from Amazon Linux machines, the ownership was NOT reassigned to `root`. I have no idea how inconsistent the behavior might be between different Linux distros? TBD?

We'll probably keep hitting this issue in the future, check out "Other fixes considered" in my writeup up top. This was the simplest change that didn't sacrifice security, so I went with it.

PR #409, opened 2022-11-01
Update submodules & Fix S3 Tests

waahm7:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/test/test_s3.py
+++ b/test/test_s3.py
@@ -242,6 +242,12 @@ class S3RequestTest(NativeResourceTest):
                 on_headers=self._on_request_headers,
                 on_body=self._on_request_body)
             finished_futures.append(s3_request.finished_future)
+            shutdown_event = s3_request.shutdown_event
+            s3_request = None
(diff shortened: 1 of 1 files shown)


test/test_s3.py
```diff
@@ -240,6 +242,7 @@ def test_put_object_multiple_times(self):
                 on_headers=self._on_request_headers,
                 on_body=self._on_request_body)
             finished_futures.append(s3_request.finished_future)
+            del s3_request
```
TingDaoK, line 245: why this is not released with as it goes out of scope?

graebm, line 245:
```suggestion
            # request keeps connection alive. delete pointer so connection can shut down
            del s3_request
```

graebm (approved):
fix & ship

test/test_s3.py
waahm7: I had a discussion with Mike and Python has a weird behavior where it is not guaranteed that something will delete when it goes out of scope.

PR #411, opened 2022-11-08
WebSocket part 1 - connect

graebm:
WIP bindings for aws-c-http's [websocket.h](https://github.com/awslabs/aws-c-http/blob/390114bc3e3cb0a53ce57accc4f619d704cb885c/include/aws/http/websocket.h)

This PR contains `websocket.connect_client(...)`

Future PRs will contain: sending frames, receiving frames, manually closing the connection.

This PR is being merged to a `websocket` feature branch, and TODOs will be addressed before it's merged to `main`

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #414, opened 2022-11-26
WebSocket part 2 - shutdown

graebm:
See part 1 here: https://github.com/awslabs/aws-crt-python/pull/411

This PR contains:
- `on_connection_shutdown` is hooked up
- `WebSocket.close()` is hooked up

### DIFF
--- a/awscrt/websocket.py
+++ b/awscrt/websocket.py
@@ -14,7 +14,6 @@ from awscrt.io import ClientBootstrap, TlsConnectionOptions, SocketOptions
 from dataclasses import dataclass
 from enum import IntEnum
 import sys
-from threading import Event
 from typing import Callable, Sequence, Tuple
 
 
@@ -199,13 +198,12 @@ class WebSocket(NativeResource):
     Use :meth:`connect()` to establish a new client connection.
     """
 
-    def __init__(self, binding, shutdown_event):
+    def __init__(self, binding):
         # Do not init a WebSocket directly, use websocket.connect()
         super().__init__()
         self._binding = binding
-        self._shutdown_event = shutdown_event
 
-    def close(self) -> Event:
+    def close(self):
         """Close the WebSocket asynchronously.
 
         You should call this when you are done with a healthy WebSocket,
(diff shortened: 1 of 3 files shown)


awscrt/websocket.py
```diff
@@ -259,7 +277,18 @@ def _on_connection_setup(
...
+            cbdata.exception = awscrt.exceptions.from_code(error_code)
+
+        # Do not let exceptions from the user's callback bubble up any further.
+        try:
+            self._on_connection_shutdown_cb(cbdata)
+        except Exception:
+            print("Exception in WebSocket on_connection_shutdown callback", file=sys.stderr)
+            sys.excepthook(*sys.exc_info())
+
+        self._shutdown_event.set()
```
TingDaoK, line 291: Why not use future and just have the exceptions raised from the future if happens? So that, we don't need to have a callback for the exception and the event for wait.

graebm: tldr: I think it's still a good idea to have a shutdown callback, that they can pass along with every other callback in the `websocket.connect()` function

Longer explanation:
I'm trying to avoid the classic creation-race-condition we keep running into when we mix callback-based APIs with stuff that's more synchronous in nature.

That's why `connect()` doesn't return a `Future[Websocket]`. The user can only get their hands on the websocket via the `on_connection_setup(...)` callback.

So if we only had a `shutdown_future`, but the user wanted to react to shutdown in an event-based way with a callback (instead of synchronously waiting on `shutdown_future.result()`) their code would need to look like this:
```py
class MyWebsocketHandler:
    def on_connection_setup(self, setup_data):
        self.websocket = setup_data.websocket
        # set shutdown callback here
        self.websocket.shutdown_future.add_done_callback(self.on_connection_shutdown)

    def on_connection_shutdown(self, shutdown_data):
        ...

    def on_incoming_frame_payload(self, incoming_data):
        ...


def main():
    ws_handler = MyWebsocketHandler()
    # but all my other callbacks are set here
    awscrt.websocket.connect(host, port,
        on_connection_setup=ws_handler.on_connection_setup,
        on_incoming_frame_payload=ws_handler.on_incoming_frame_payload)
```

I think the user would prefer to pass all their callbacks in one place, like:
```py
def main():
    ws_handler = MyWebsocketHandler()
    # but all my other callbacks are set here
    awscrt.websocket.connect(host, port,
        on_connection_setup=ws_handler.on_connection_setup,
        on_connection_shutdown=ws_handler.on_connection_shutdown,
        on_incoming_frame_payload=ws_handler.on_incoming_frame_payload)
```

graebm: I flip-flopped on whether to return `Event`, `Future`, or `None` from `close()`....

we just chatted and you mentioned how offering multiple ways of learning about something can confuse the user. I agree. Let's keep it simple and not return anything from `close()`. We could always add something later

TingDaoK (approved):
🚢

PR #415, opened 2022-11-30
Raise AwsCrtError from C, instead of basic RuntimeError.

graebm:
**Issue:**
The exceptions raised by sync vs async C code are different.

Sync C code always raises a  generic `RuntimeError` with everything in a long text description, like:
`RuntimeError("2066 (AWS_ERROR_HTTP_WEBSOCKET_CLOSE_FRAME_SENT): Websocket has sent CLOSE frame, no more data will be sent.")`

Async code will raise a nicer `AwsCrtError`, with separate `name`, `code`, and `message` properties, like:
`AwsCrtError(name="AWS_ERROR_HTTP_WEBSOCKET_CLOSE_FRAME_SENT", message="Websocket has sent CLOSE frame, no more data will be sent.", code=2066)`

**Background:**
It's not easy for our C code to use a class like `AwsCrtError` that is defined in our Python code.

**Description of changes:**
- Sync C code raises `AwsCrtError` instead of `RuntimeError`
- Change `AwsCrtError`'s to inherit from `RuntimeError` instead of `Exception`
  - This if for backwards compatibility, if anyone wrote code to catch a `RuntimeException` they'll still catch the `AwsCrtError`
- Do some trickery to initialize the C library with stuff from our Python code

### DIFF
--- a/awscrt/_c_lib_importer.py
+++ b/awscrt/_c_lib_importer.py
@@ -1,0 +1,29 @@
+"""
+Our python files MUST import _awscrt like this:
+```
+from awscrt._c_lib_importer import _awscrt
+```
+
+to ensure that _awscrt.init_c_library_with_python_stuff(...)
+is called before any other C function.
+
+Unfortunately, we can get circular dependencies when the "python stuff" is defined
+in another file that also needs to import _c_lib_importer.
+If this happens to you, hack around it like this:
+```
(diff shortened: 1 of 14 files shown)


graebm:
Note to self: Try defining AwsCrtError in _awscrt, instead of having a circular dependency

PR #416, opened 2022-12-03
WebSocket part 3 - send and receive

graebm:
See part 1 here: https://github.com/awslabs/aws-crt-python/pull/411
See part 2 here: https://github.com/awslabs/aws-crt-python/pull/414

This PR contains:
- WebSocket.send_frame() is hooked up
- The `on_incoming_frame` callbacks are hooked up

Unrelated changes:
- `scripts/make-docs` is now python instead of bash because I hate bash
- Update devguide with better instructions for setting virtual environment


awscrt/websocket.py
```diff
@@ -14,26 +30,25 @@
 from dataclasses import dataclass
 from enum import IntEnum
 import sys
-from typing import Callable, Sequence, Tuple
+from typing import Callable, Optional, Sequence, Tuple, Union
 
 
 def connect(
     *,
```
SergeyRyabinin, line 37: you might really want kwargs here now...

awscrt/websocket.py
```diff
@@ -261,11 +337,12 @@ def _on_connection_setup(
 
         # Do not let exceptions from the user's callback bubble up any further.
         try:
-            self._on_connection_setup_cb(cbdata)
+            if self._on_connection_setup_cb is not None:
```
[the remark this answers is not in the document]
graebm: WOOPS

great catch
this was one of my last edits, I forgot the if-check on all the callbacks, but this one shouldn't have ever had one

awscrt/websocket.py
graebm: Not sure what you mean...
Do you mean that I should take `**kwargs`?

If so, I agree, and I'm doing one better by using the Python 3.0+ feature: [Keyword-Only Arguments](https://peps.python.org/pep-3102/)

SergeyRyabinin: I missed the * on the first line, my bad.
Thanks for the link!

PR #418, opened 2022-12-06
Mqtt5 canary

sbSteveK:
Add short-term MQTT5 Canary


test/mqtt5_canary.py
```diff
@@ -0,0 +1,282 @@
...
+        client.initial_start()
+
+    for client in clients:
+        client.subscribe()
+
+    while time.time() < time_end:
+        if time.time() >= time_next_operation:
+            time_next_operation += tpsdelay
+            for client in clients:
+                client.random_operation()
```
xiazhvera, line 271: A minor issue, the behavior for random operation here is different from the one in aws-c-mqtt. In aws-c-mqtt, it runs 1 operation/tpsdelay, while in python it runs 1 operation/client/tpsdelay. 
I think either behavior works, just need to note that it actually runs N*tps operations per seconds in total.

xiazhvera (approved):

test/mqtt5_canary.py
sbSteveK: Ah! Great catch! Fixed to run the correct TPS.

PR #422, opened 2022-12-08
handle null ack packets in callbacks

sbSteveK:
Ack packets can potentially be NULL and must be handled in bindings.


graebm (approved):
simplify & ship

PR #427, opened 2022-12-14
Add Option to configure Hybrid PQ TLS Cipher Preferences

alexw91:
*Issue #, if available:*
N/A

*Description of changes:*
Add Option to configure Hybrid PQ TLS Cipher Preferences. 

Note: I don't write a lot of Python, and this is the first time I'm writing Python-to-C translation code, so please take an extra close look when reviewing to make sure I didn't make any simple mistakes. 

### DIFF
--- a/awscrt/io.py
+++ b/awscrt/io.py
@@ -10,7 +10,7 @@ Long-running event-loop threads are used for concurrency.
 
 import _awscrt
 from awscrt import NativeResource
-from enum import Enum, IntEnum
+from enum import IntEnum
 import threading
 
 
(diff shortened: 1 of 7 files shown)


awscrt/io.py
```diff
@@ -250,6 +250,14 @@ class TlsVersion(IntEnum):
     DEFAULT = 128  #:
 
 
+class TlsCipherPref(Enum):
```
graebm, line 253: I'd just make this an IntEnum like all our other enums

awscrt/io.py
```diff
@@ -250,6 +250,14 @@ class TlsVersion(IntEnum):
...
+    DEFAULT = (0)
+    PQ_TLSv1_0_2021_05 = (6)
+
+    def __init__(self, tlsCipherEnumVal):
+        self.isSupported = _awscrt.is_tls_cipher_supported(tlsCipherEnumVal)
```
graebm, line 258: I'd do this as a member function, instead of an attribute. It's transparent, the action in python maps to the action in C. Also, it lets you just make this a normal IntEnum, you don't have to mess with the `__init__` function

also, our python API uses snake_case, not camelCase

also, need to include a docstring or it won't show up in the API docs

```
    def is_supported(self):
        """Return whether this Cipher Preference is available in the underlying TLS implementation"""
        return _awscrt.is_tls_cipher_supported(self.value)

```

awscrt/io.py
```diff
@@ -250,6 +250,14 @@ class TlsVersion(IntEnum):
     DEFAULT = 128  #:
 
 
+class TlsCipherPref(Enum):
+    DEFAULT = (0)
```
graebm, line 254: stuff needs to have SOME kind of docstring or it won't show up in the [API docs](https://awslabs.github.io/aws-crt-python/api/io.html#awscrt.io.TlsVersion)

you need to add a """ triple-quoted docstring, or at the very least a `#:` or the enum value won't even show up

you can see what the docs will look like by locally running `./scripts/make-docs.py`, then look at `docs/index.html` in your browser

source/io.c
```diff
@@ -83,6 +83,18 @@ PyObject *aws_py_is_alpn_available(PyObject *self, PyObject *args) {
...
 }
 
+PyObject *aws_py_is_tls_cipher_supported(PyObject *self, PyObject *args) {
+    (void)self;
+    (void)args;
+
+    int cipher_pref = 0;
+
+    if (!PyArg_ParseTuple(args, "i", &cipher_pref)) {
+        return PyErr_AwsLastError();
```
graebm, line 93:
```suggestion
        return NULL;
```

Py_ArgParseTuple() is a python-c-api function, it doesn't raise an AWS error

the way to report an exception in a python-c-api function is to set the current exception, and return NULL

PyArg_ParseTuple has already set the exception in this case, so we just need to return NULL

test/test_crypto.py
```diff
@@ -75,6 +76,9 @@ def test_md5_iterated(self):
         expected = b'\x90\x01\x50\x98\x3c\xd2\x4f\xb0\xd6\x96\x3f\x7d\x28\xe1\x7f\x72'
         self.assertEqual(expected, digest)
 
+    def test_is_tls_cipher_supported(self):
```
graebm, line 79: trivial: put this test in `test_io.py` since it's testing something from `io.py`

this file is for testing stuff from `crypto.py`

test/test_http_client.py
```diff
@@ -77,6 +77,8 @@ def _stop_server(self):
     def _new_client_connection(self, secure, proxy_options=None):
         if secure:
             tls_ctx_opt = TlsContextOptions()
+            if (TlsCipherPref.PQ_TLSv1_0_2021_05.isSupported):
+                tls_ctx_opt.cipher_pref = TlsCipherPref.PQ_TLSv1_0_2021_05
```
graebm, line 81: let's test this explicitly, instead of sneaking it into our vanilla tests

I'd propose doing it something like:

add an extra param like:
`def _new_client_connection(self, secure, proxy_options=None, cipher_pref=TlsCipherPref.DEFAULT):`

add an explicit test (that's skipped if the cipher isn't supported) something like:
```
    @unittest.skipIf(not TlsCipherPref.PQ_TLSv1_0_2021_05.is_supported())
    def test_connect_pq_tlsv1_0_2021_05(self):
        self._test_connect(secure=True, cipher_pref=TlsCipherPref.PQ_TLSv1_0_2021_05)
```

awscrt/io.py
```diff
@@ -269,6 +277,7 @@ class TlsContextOptions:
         'min_tls_ver',
```
graebm, line 277: need to add cipher_pref into the list of "Attributes", in the docstring for TlsContextOptions

graebm (approved):

PR #439, opened 2023-01-26
Remove future from operation statistics

TwistedTwigleg:
*Description of changes:*

Adjusts operation statistics getting functions in both MQTT311 and MQTT5 to return the statistics directly rather than via a future.

_____________

### DIFF
--- a/source/mqtt5_client.c
+++ b/source/mqtt5_client.c
@@ -2062,21 +2062,11 @@ done:
         return result;
     }
 
-    if (result) {
-        Py_XDECREF(result);
-    }
-    if (incomplete_operation_count_obj) {
-        Py_XDECREF(incomplete_operation_count_obj);
-    }
-    if (incomplete_operation_size_obj) {
-        Py_XDECREF(incomplete_operation_size_obj);
-    }
-    if (unacked_operation_count_obj) {
-        Py_XDECREF(unacked_operation_count_obj);
-    }
-    if (unacked_operation_size_obj) {
-        Py_XDECREF(unacked_operation_size_obj);
-    }
+    Py_XDECREF(result);
+    Py_XDECREF(incomplete_operation_count_obj);
+    Py_XDECREF(incomplete_operation_size_obj);
+    Py_XDECREF(unacked_operation_count_obj);
+    Py_XDECREF(unacked_operation_size_obj);
 
     return NULL;
 }
--- a/source/mqtt5_client.c
+++ b/source/mqtt5_client.c
@@ -2026,10 +2026,10 @@ PyObject *aws_py_mqtt5_client_get_stats(PyObject *self, PyObject *args) {
     aws_mqtt5_client_get_stats(client->native, &stats);
 
     result = PyTuple_New(4);
-    PyTuple_SetItem(result, 0, (unsigned long long)stats.incomplete_operation_count);
-    PyTuple_SetItem(result, 1, (unsigned long long)stats.incomplete_operation_size);
-    PyTuple_SetItem(result, 2, (unsigned long long)stats.unacked_operation_count);
-    PyTuple_SetItem(result, 3, (unsigned long long)stats.unacked_operation_size);
+    PyTuple_SET_ITEM(result, 0, PyLong_FromUnsignedLongLong((unsigned long long)stats.incomplete_operation_count));
+    PyTuple_SET_ITEM(result, 1, PyLong_FromUnsignedLongLong((unsigned long long)stats.incomplete_operation_size));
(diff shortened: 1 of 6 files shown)


source/mqtt5_client.c
```diff
@@ -2018,34 +2017,28 @@ PyObject *aws_py_mqtt5_client_get_stats(PyObject *self, PyObject *args) {
...
-        /* K */ (unsigned long long)stats.incomplete_operation_size,
-        /* K */ (unsigned long long)stats.unacked_operation_count,
-        /* K */ (unsigned long long)stats.unacked_operation_size);
-    if (!result) {
-        PyErr_WriteUnraisable(PyErr_Occurred());
-        goto done;
-    }
-
+    result = PyTuple_New(4);
+    PyTuple_SET_ITEM(result, 0, PyLong_FromUnsignedLongLong((unsigned long long)stats.incomplete_operation_count));
```
graebm, line 2029: need to check results of PyTuple_New() and PyLongFromUnsignedLongLong() calls. Each may  report failure

source/mqtt5_client.c
```diff
@@ -2018,34 +2017,66 @@ PyObject *aws_py_mqtt5_client_get_stats(PyObject *self, PyObject *args) {
...
-
     if (success) {
-        Py_RETURN_NONE;
+        return result;
+    }
+
+    if (result) {
+        Py_XDECREF(result);
     }
+    if (incomplete_operation_count_obj) {
```
sbSteveK, line 2068: You don't need to check for NULL on Py_XDECREF

sbSteveK (approved):

source/mqtt5_client.c
```diff
@@ -2018,34 +2017,56 @@ PyObject *aws_py_mqtt5_client_get_stats(PyObject *self, PyObject *args) {
...
+    result = PyTuple_New(4);
     if (!result) {
-        PyErr_WriteUnraisable(PyErr_Occurred());
         goto done;
     }
 
+    incomplete_operation_count_obj = PyLong_FromUnsignedLongLong((unsigned long long)stats.incomplete_operation_count);
+    if (!incomplete_operation_count_obj) {
+        goto done;
+    }
```
graebm, line 2040: See my next comment explaining WHY this should probably look like:

```
num_obj = PyLong_FromUnsignedLongLong(stats.incomplete_operation_count);
if (!num_obj) {
    goto done;
}
PyTuple_SET_ITEM(result, 0, num_obj); /* steals reference to num_obj */

num_obj = PyLong_FromUnsignedLongLong(stats.incomplete_operation_size);
if (!num_obj) {
    goto done;
}
PyTuple_SET_ITEM(result, 1, num_obj); /* steals reference to num_obj */
...
```

or maybe even simpler:

```
PyTuple_SET_ITEM(result, 0, PyLong_FromUnsignedLongLong(stats.incomplete_operation_count));
if (PyTuple_GET_ITEM(result, 0) == NULL) {
    goto done;
}

PyTuple_SET_ITEM(result, 1, PyLong_FromUnsignedLongLong(stats.incomplete_operation_size));
if (PyTuple_GET_ITEM(result, 1) == NULL) {
    goto done;
}
...
```

source/mqtt5_client.c
```diff
@@ -2018,34 +2017,56 @@ PyObject *aws_py_mqtt5_client_get_stats(PyObject *self, PyObject *args) {
...
 done:
+    if (success) {
+        return result;
+    }
 
     Py_XDECREF(result);
+    Py_XDECREF(incomplete_operation_count_obj);
+    Py_XDECREF(incomplete_operation_size_obj);
+    Py_XDECREF(unacked_operation_count_obj);
+    Py_XDECREF(unacked_operation_size_obj);
```
graebm, line 2069: DON'T decref these.

PyTuple_SET_ITEM() "steals" a reference to the number object (adds it to the tuple without altering their refcount, essentially transferring ownership to the tuple)

so with this code, if item 2/4 fails, then `Py_XDECREF(result);` will make the `result` get destroyed, which will in turn decref everything inside it, bringing the refcount of`incomplete_operation_count_obj` to 0, which destroys it, but now the `incomplete_operation_count_obj` variable is pointing to garbage memory, and `Py_XDECREF(incomplete_operation_count_obj)` will be like "this is not null, so decref it" which is use-after-free, which is a segfault in the best case, and undefined behavior in the worst case

TwistedTwigleg: That sounds bad! I adjusted it to avoid this 👍

graebm (approved):

TwistedTwigleg:

PR #441, opened 2023-02-05
Patch 1

minegit:
*Issue #, if available:*

*Description of changes:*
Adding **kwargs to allow the user to pass objects like flask context, any other meta data to the callback method.

### DIFF
--- a/awscrt/mqtt.py
+++ b/awscrt/mqtt.py
@@ -476,7 +476,7 @@ class Connection(NativeResource):
 
         return future
 
-    def subscribe(self, topic, qos, callback=None):
+    def subscribe(self, topic, qos, callback=None, **kwargs):
         """Subscribe to a topic filter (async).
 
         The client sends a SUBSCRIBE packet and the server responds with a SUBACK.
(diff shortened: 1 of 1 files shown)


bretambrose (commented):
If you want additional context available within your callbacks, we recommend using lambda capture mechanics when defining the callback.  Using kwargs as an opaque user data passing mechanism over async boundaries is not a pattern I think we want to support or encourage.

minegit:
Hi @bretambrose can you give me any example in python or any link where I can understand this mechanism. 
Because we require some context to be passed for the callback to process the message and we are stuck here .

bretambrose:
> Hi @bretambrose can you give me any example in python or any link where I can understand this mechanism. Because we require some context to be passed for the callback to process the message and we are stuck here .

Both lambdas and inner functions work fine.  Here's an example using an inner function.  The extended callback (with context parameter) is called with the additional context that is captured by the inner function's definition at subscribe time.

```
def on_message_received_with_context(topic, payload, dup, qos, retain, context):
    print("Received message from topic '{}': {}".format(topic, payload))
    print("Captured context: {}".format(context))

... (in other function)

    context = "Hello there"
    def on_message_received(topic, payload, dup, qos, retain):
        on_message_received_with_context(topic, payload, dup, qos, retain, context)

    # Subscribe
    print("Subscribing to topic '{}'...".format(message_topic))
    subscribe_future, packet_id = mqtt_connection.subscribe(
        topic=message_topic,
        qos=mqtt.QoS.AT_LEAST_ONCE,
        callback=on_message_received)

```

PR #442, opened 2023-02-08
MQTT311 extra connection callbacks

TwistedTwigleg:
*Description of changes:*

Adds additional connection callbacks to MQTT311. Specifically, adds `onConnectionSuccess`, `onConnectionFailure`, and `onConnectionClosed` These callbacks mirror those in V1.

____________

### DIFF
--- a/source/mqtt_client_connection.c
+++ b/source/mqtt_client_connection.c
@@ -163,22 +163,35 @@ static void s_on_connection_closed(
 
     struct mqtt_connection_binding *py_connection = userdata;
 
+    fprintf(stderr, "\n TEST 01 \n");
+
     PyGILState_STATE state;
     if (aws_py_gilstate_ensure(&state)) {
         return; /* Python has shut down. Nothing matters anymore, but don't crash */
     }
 
+    fprintf(stderr, "\n TEST 02 \n");
+
     /* Ensure that python class is still alive */
-    PyObject *self = PyWeakref_GetObject(py_connection->self_proxy); /* borrowed reference */
-    if (self != Py_None) {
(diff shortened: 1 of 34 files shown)


source/mqtt_client_connection.c
```diff
@@ -53,6 +53,9 @@ struct mqtt_connection_binding {
 
     /* Dependencies that must outlive this */
     PyObject *client;
+
+    /* Reference count - needed to keep the connection alive even if the Python object is destroyed */
+    struct aws_atomic_var ref_count;
```
TwistedTwigleg, line 58: Note: I had to add reference counting in C here, because it is possible for a Python program to be waiting on the disconnect to finish, and then as soon as it finishes it shuts down. Prior to this change, the Python object's cleanup in C was tied 1:1 with the cleanup of the Python object, and so what would happen is the Python object would be cleaned right as the `on_closed` callback is invoked (as `on_closed` happens AFTER the disconnect) and this would lead to unavoidable segfaults!

Using a quick reference count fixes this issue by allowing the C object to live longer than the Python one, allowing the code to tell it to stay alive until the `on_stopped` is called. I also added this reference counting to the connection callback/events as well.

TwistedTwigleg:
Note: Once merged, a follow-up PR will need to be made in the Python V2 SDK to expose the new callbacks, as the MQTT311 builder in the V2 SDK currently doesn't directly pass the `kwargs` along.

source/mqtt_client_connection.c
TwistedTwigleg: After discussion with Vera, we are going to see if it is possible to work around this issue and not use reference counting in C.

xiazhvera (approved):

TwistedTwigleg:

PR #446, opened 2023-02-16
Allow passing None for the PKCS11 user pin

TwistedTwigleg:
*Description of changes:*

The documentation for the PKCS11 `create_client_with_mtls_pkcs11()` function states that you can pass `None` as a valid input, but this is not allowed because `user_pin` is not set to `None` by default. When using Python with typing, this causes problems.

This PR fixes that by making the `user_pin` optional by setting it to `null` by default. It also adjusts the C code to check if the value is `null` and, if it is, it passes a zeroed out `aws_byte_cursor`.

___________

### DIFF
--- a/awscrt/io.py
+++ b/awscrt/io.py
@@ -374,7 +374,7 @@ class TlsContextOptions:
     @staticmethod
     def create_client_with_mtls_pkcs11(*,
                                        pkcs11_lib: 'Pkcs11Lib',
-                                       user_pin: str,
+                                       user_pin: str = None,
                                        slot_id: int = None,
                                        token_label: str = None,
                                        private_key_label: str = None,
@@ -390,7 +390,7 @@ class TlsContextOptions:
             pkcs11_lib (Pkcs11Lib): Use this PKCS#11 library
 
             user_pin (str): User PIN, for logging into the PKCS#11 token.
-                Pass `None` to log into a token with a "protected authentication path".
+                Pass `None` or do not specify to log into a token with a "protected authentication path".
 
             slot_id (Optional[int]): ID of slot containing PKCS#11 token.
                 If not specified, the token will be chosen based on other criteria (such as token label).
--- a/awscrt/io.py
+++ b/awscrt/io.py
@@ -12,6 +12,7 @@ import _awscrt
 from awscrt import NativeResource
 from enum import IntEnum
 import threading
+from typing import Union
 
 
 class LogLevel(IntEnum):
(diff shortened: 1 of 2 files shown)


awscrt/io.py
```diff
@@ -374,7 +374,7 @@ def create_client_with_mtls(cert_buffer, key_buffer):
     @staticmethod
     def create_client_with_mtls_pkcs11(*,
                                        pkcs11_lib: 'Pkcs11Lib',
-                                       user_pin: str,
+                                       user_pin: str = None,
```
graebm, line 377: Subtlety: While users **can** pass None, we don't want that to be the default value. Passing `None` to "log into the token with a protected authentication path" is an uncommon use-case that people should be explicit about

```suggestion
                                       user_pin: Union[str, None],
```

I feel like `Union[str, None]`

you'll also need `from typing import Union` at the top of the file

Newer versions of python would have simply let us write this as `str|None` but in Python 3.7 we need to write it as `Union[str, None]`

awscrt/io.py
```diff
@@ -390,7 +390,7 @@ def create_client_with_mtls_pkcs11(*,
             pkcs11_lib (Pkcs11Lib): Use this PKCS#11 library
 
             user_pin (str): User PIN, for logging into the PKCS#11 token.
-                Pass `None` to log into a token with a "protected authentication path".
+                Pass `None` or do not specify to log into a token with a "protected authentication path".
```
graebm, line 393: revert

source/io.c
```diff
@@ -501,9 +501,16 @@ PyObject *aws_py_client_tls_ctx_new(PyObject *self, PyObject *args) {
...
+            AWS_ZERO_STRUCT(user_pin_cursor);
+        }
+
         struct aws_tls_ctx_pkcs11_options pkcs11_options = {
             .pkcs11_lib = pkcs11_lib,
-            .user_pin = aws_byte_cursor_from_array(pkcs11_user_pin, pkcs11_user_pin_len),
+            .user_pin = user_pin_cursor,
             .slot_id = has_slot_id ? &slot_id_value : NULL,
             .token_label = aws_byte_cursor_from_array(pkcs11_token_label, pkcs11_token_label_len),
             .private_key_object_label = aws_byte_cursor_from_array(pkcs11_priv_key_label, pkcs11_priv_key_label_len),
```
graebm, line 516: revert

if `pkcs11_user_pin` is NULL then `pkcs11_user_pin_len` is 0 and the old code handles that fine. I agree that your change is more explicit, but we're already taking the same approach of "just calling aws_byte_cursor_from_array() is fine" with all the other strings/cursors in this options struct

graebm (approved):

TwistedTwigleg:

PR #450, opened 2023-03-08
Do not decref the result object twice

nforro:
The result object is decref'd immediately after the call to `PyObject_CallFunction()`, do not decref it again.

This fixes `Fatal Python error: deallocating None` when running the test suite with the module built for Python 3.11 on Fedora.


graebm (approved):
Thanks!

Yikes. This bug has been here a long time. Embarrassing. I just searched the rest of the codebase for places we decref `result`, and this appears to be the only place we accidentally do it twice

PR #454, opened 2023-03-13
Make it possible to use OpenSSL on Unix

nforro:
This patch makes it possible to set USE_OPENSSL environment variable to build and link the module with OpenSSL rather than AWS-LC on Unix.

Note: aws-c-cal already accepts USE_OPENSSL CMake option, so setting the environment variable affects it as well.

### DIFF
--- a/crt/CMakeLists.txt
+++ b/crt/CMakeLists.txt
@@ -28,7 +28,11 @@ if(UNIX AND NOT APPLE)
     set(DISABLE_GO ON) # Build without using Go, we don't want the extra dependency
     set(DISABLE_PERL ON) # Build without using Perl, we don't want the extra dependency
     set(BUILD_LIBSSL OFF) # Don't need libssl, only need libcrypto
-    add_subdirectory(aws-lc)
+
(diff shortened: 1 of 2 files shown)


graebm:
Curious, what's the motivation behind this? What problem are you trying to solve?

nforro:
I'm packaging aws-crt-python for Fedora, and the package needs to comply with [Fedora Crypto Policies](https://docs.fedoraproject.org/en-US/packaging-guidelines/CryptoPolicies/). That can be conveniently achieved by linking with OpenSSL.

Another option would be to make AWS-LC compliant with the policies, but I don't know how realistic that is.

graebm:
Cool, yeah, just making sure there's a good motivation before we increase the surface area of things to support.

Talked about it with the team. We want to use a different string, and want to add CI to ensure that it works. We'll get something together ASAP

nforro:
> Talked about it with the team. We want to use a different string, and want to add CI to ensure that it works. We'll get something together ASAP

Great, thanks!

nforro:
Closing in favor of #456.

PR #456, opened 2023-03-14
Add AWS_CRT_BUILD_USE_SYSTEM_LIBCRYPTO

TingDaoK:
original: https://github.com/awslabs/aws-crt-python/pull/454

- Use `AWS_CRT_BUILD_USE_SYSTEM_LIBCRYPTO` as the env var name instead
- Added CI

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -102,6 +102,18 @@ jobs:
           aws s3 cp s3://aws-crt-test-stuff/ci/${{ env.BUILDER_VERSION }}/linux-container-ci.sh ./linux-container-ci.sh && chmod a+x ./linux-container-ci.sh
           ./linux-container-ci.sh ${{ env.BUILDER_VERSION }} aws-crt-${{ env.LINUX_BASE_IMAGE }} build -p ${{ env.PACKAGE_NAME }} --compiler=${{ matrix.compiler }}
 
+  use-openssl:
+    runs-on: ubuntu-20.04 # latest
+    steps:
+      - name: Build ${{ env.PACKAGE_NAME }}
+        env:
+          AWS_CRT_BUILD_USE_SYSTEM_LIBCRYPTO: 1
+        runs: |
+          python3 -c "from urllib.request import urlretrieve; urlretrieve('${{ env.BUILDER_HOST }}/${{ env.BUILDER_SOURCE }}/${{ env.BUILDER_VERSION }}/builder.pyz?run=${{ env.RUN }}', 'builder')"
+          chmod a+x builder
+          ./builder build -p ${{ env.PACKAGE_NAME }}
+
+
 
   windows:
     runs-on: windows-2022 # latest
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -102,7 +102,7 @@ jobs:
           aws s3 cp s3://aws-crt-test-stuff/ci/${{ env.BUILDER_VERSION }}/linux-container-ci.sh ./linux-container-ci.sh && chmod a+x ./linux-container-ci.sh
           ./linux-container-ci.sh ${{ env.BUILDER_VERSION }} aws-crt-${{ env.LINUX_BASE_IMAGE }} build -p ${{ env.PACKAGE_NAME }} --compiler=${{ matrix.compiler }}
 
-  use-openssl:
+  use-system-libcrypto:
     runs-on: ubuntu-20.04 # latest
     steps:
       - name: Build ${{ env.PACKAGE_NAME }}
@@ -113,6 +113,15 @@ jobs:
           chmod a+x builder
           ./builder build -p ${{ env.PACKAGE_NAME }}
 
+          # assert it's linked against the system's libcrypto.so
+          AWSCRT_PATH=`python3 -c "import _awscrt; print(_awscrt.__file__)"`
+          printf "AWSCRT_PATH: $AWSCRT_PATH\n"
+
+          LINKED_AGAINST=`ldd $AWSCRT_PATH`
+          printf "LINKED AGAINST:\n$LINKED_AGAINST\n"
+
+          USES_LIBCRYPTO_SO=`echo "$LINKED_AGAINST" | grep 'libcrypto*.so' | head -1`
+          test -n "$USES_LIBCRYPTO_SO"
 
 
   windows:
@@ -140,6 +149,10 @@ jobs:
   openbsd:
     runs-on: ubuntu-latest # macos would be faster. TODO: diagnose why some tests are failing
     steps:
+      # Cannot use builder to checkout as OpenBSD doesn't ship git in the base install
+      - uses: actions/checkout@v3
+        with:
+          submodules: true
       - name: Build ${{ env.PACKAGE_NAME }} + consumers
         uses: cross-platform-actions/action@v0.10.0
         with:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -112,8 +112,6 @@ jobs:
           python3 -c "from urllib.request import urlretrieve; urlretrieve('${{ env.BUILDER_HOST }}/${{ env.BUILDER_SOURCE }}/${{ env.BUILDER_VERSION }}/builder.pyz?run=${{ env.RUN }}', 'builder')"
           chmod a+x builder
           ./builder build -p ${{ env.PACKAGE_NAME }}
-
-    steps:
       - name: Assert libcrypto.so used
         run: |
           # assert it's linked against the system's libcrypto.so
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -108,7 +108,7 @@ jobs:
       - name: Build ${{ env.PACKAGE_NAME }}
         env:
           AWS_CRT_BUILD_USE_SYSTEM_LIBCRYPTO: 1
-        runs: |
+        run: |
           python3 -c "from urllib.request import urlretrieve; urlretrieve('${{ env.BUILDER_HOST }}/${{ env.BUILDER_SOURCE }}/${{ env.BUILDER_VERSION }}/builder.pyz?run=${{ env.RUN }}', 'builder')"
           chmod a+x builder
           ./builder build -p ${{ env.PACKAGE_NAME }}
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -35,6 +35,9 @@ jobs:
           - cp310-cp310
           - cp311-cp311
     steps:
+    - uses: actions/checkout@v3
+      with:
+        submodules: recursive
     # Only aarch64 needs this, but it doesn't hurt anything
     - name: Install qemu/docker
       run: docker run --rm --privileged multiarch/qemu-user-static --reset -p yes
@@ -53,6 +56,9 @@ jobs:
         image:
           - raspbian-bullseye
     steps:
+    - uses: actions/checkout@v3
+      with:
+        submodules: recursive
     # set arm arch
     - name: Install qemu/docker
       run: docker run --rm --privileged multiarch/qemu-user-static --reset -p yes
@@ -73,6 +79,9 @@ jobs:
           - opensuse-leap
           - rhel8-x64
     steps:
+    - uses: actions/checkout@v3
+      with:
+        submodules: recursive
         # We can't use the `uses: docker://image` version yet, GitHub lacks authentication for actions -> packages
     - name: Build ${{ env.PACKAGE_NAME }}
       run: |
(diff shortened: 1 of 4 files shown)


setup.py
```diff
@@ -121,6 +121,9 @@ def get_cmake_path():
     raise Exception("CMake must be installed to build from source.")
 
 
+use_openssl = os.environ.get('AWS_CRT_BUILD_USE_SYSTEM_LIBCRYPTO', False)
```
graebm, line 124: Let's try and use the term "use system libcrypto" as much as possible, vs the kinda misleading "use openssl" term

Also, this file tends to uses functions over global variables, so let's stick with that

Also, this line weirdly mixes strings and booleans (`use_openssl` could end up as False or "" or "0" or "1")

maybe:
```
def using_system_libcrypto():
    # if we're going ifdef style
    return 'AWS_CRT_BUILD_USE_SYSTEM_LIBCRYPTO' in os.environ
    # or if we're going set-to-1 style   
    return os.environ.get(AWS_CRT_BUILD_USE_SYSTEM_LIBCRYPTO) == "1"
```

crt/CMakeLists.txt
```diff
@@ -28,7 +28,11 @@ if(UNIX AND NOT APPLE)
     set(DISABLE_GO ON) # Build without using Go, we don't want the extra dependency
     set(DISABLE_PERL ON) # Build without using Perl, we don't want the extra dependency
     set(BUILD_LIBSSL OFF) # Don't need libssl, only need libcrypto
```
graebm, line 30: move these `set()` calls inside the `if (NOT USE_OPENSSL)`

they exist to customize how aws-lc is built

setup.py
```diff
@@ -305,10 +313,13 @@ def awscrt_ext():
...
-        # this prevents weird crashes if an application also ends up using
-        # libcrypto.so from the system's OpenSSL installation.
-        extra_link_args += ['-Wl,--exclude-libs,libcrypto.a']
+        if using_system_libcrypto():
+            extra_link_args += ['-lcrypto']
+        else:
+            # hide the symbols from libcrypto.a
+            # this prevents weird crashes if an application also ends up using
+            # libcrypto.so from the system's OpenSSL installation.
+            extra_link_args += ['-Wl,--exclude-libs,libcrypto.a']
```
DmitriyMusatkin, line 322: can you verify that we are not leaking symbols when compiling against system libcrypto? we might want to keep symbol stripping behavior even for system crypto branch. (nm or objdump will let you dump symbols)

setup.py
```diff
@@ -121,8 +121,8 @@ def get_cmake_path():
     raise Exception("CMake must be installed to build from source.")
 
 
-use_openssl = os.environ.get('AWS_CRT_BUILD_USE_SYSTEM_LIBCRYPTO', False)
-
+def using_system_libcrypto():
+    return 'AWS_CRT_BUILD_USE_SYSTEM_LIBCRYPTO' in os.environ
```
nforro, line 125: This would be `True` for `AWS_CRT_BUILD_USE_SYSTEM_LIBCRYPTO=0` though. I think the alternative is better:

```python
return os.getenv('AWS_CRT_BUILD_USE_SYSTEM_LIBCRYPTO') == '1'
```

It is also already used for example here: https://github.com/awslabs/aws-crt-python/blob/238d38627d925e16a403d184f2b1aef495de9889/awscrt/_test.py#L109

setup.py
```diff
@@ -309,6 +317,8 @@ def awscrt_ext():
         # this prevents weird crashes if an application also ends up using
         # libcrypto.so from the system's OpenSSL installation.
         extra_link_args += ['-Wl,--exclude-libs,libcrypto.a']
+        if using_system_libcrypto():
+            extra_link_args += ['-lcrypto']
```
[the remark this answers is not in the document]
graebm: also, add it to the list of `libraries` not ~extra_link_args~
for consistency
```
        if using_system_libcrypto():
            libraries += ['crypto']
```

.github/workflows/ci.yml
```diff
@@ -96,19 +105,40 @@ jobs:
           - gcc-7
           - gcc-8
     steps:
+      - uses: actions/checkout@v3
+        with:
+          submodules: recursive
```
graebm, line 110: I'm pretty sure this is unnecessary. The only tests right now that do checkout are ones that don't use builder

(huh, looks like the new `openbsd` test does checkout ... probably unnecessary)

README.md
```diff
@@ -38,3 +41,28 @@ Please note that on Mac, once a private key is used with a certificate, that cer
 ```
 static: certificate has an existing certificate-key pair that was previously imported into the Keychain. Using key from Keychain instead of the one provided.
 ```
+
+### OpenSSL and LibCrypto (Unix only)
```
graebm, line 45: Move this up above the `Mac-Only TLS Behavior` section

This openssl stuff is relevent to building/installation, it should be close to the existing build/install steps

README.md
```diff
@@ -38,3 +41,28 @@ Please note that on Mac, once a private key is used with a certificate, that cer
...
+You can ignore all this on Windows and Apple platforms, where aws-crt-python
+uses the OS's default libraries for TLS and cryptography math.
+To simplify installation, aws-crt-python has its own copy of libcrypto.
+This lets you install a wheel from PyPI without having OpenSSL installed.
+Unix wheels on PyPI come with libcrypto statically compiled in.
+Code to build libcrypto comes from [AWS-LC](https://github.com/aws/aws-lc).
+AWS-LC's code is included in the PyPI source package, 
+and the git repository includes it as a submodule.
+If you need aws-crt-python to use the libcrypto included on your system, 
+set environment variable `AWS_CRT_BUILD_USE_SYSTEM_LIBCRYPTO=1` while building from source:
```
graebm, line 60: paragraph breaks
```suggestion
aws-crt-python does not use OpenSSL for TLS.
On Apple and Windows devices, the OS's default TLS library is used.
On Unix devices, [s2n-tls](https://github.com/aws/s2n-tls) is used.
But s2n-tls uses libcrypto, the cryptography math library bundled with OpenSSL.

You can ignore all this on Windows and Apple platforms, where aws-crt-python
uses the OS's default libraries for TLS and cryptography math.

To simplify installation, aws-crt-python has its own copy of libcrypto.
This lets you install a wheel from PyPI without having OpenSSL installed.
Unix wheels on PyPI come with libcrypto statically compiled in.
Code to build libcrypto comes from [AWS-LC](https://github.com/aws/aws-lc).
AWS-LC's code is included in the PyPI source package, 
and the git repository includes it as a submodule.

If you need aws-crt-python to use the libcrypto included on your system, 
set environment variable `AWS_CRT_BUILD_USE_SYSTEM_LIBCRYPTO=1` while building from source:
```

README.md
```diff
@@ -38,3 +41,28 @@ Please note that on Mac, once a private key is used with a certificate, that cer
...
+
+```sh
+AWS_CRT_BUILD_USE_SYSTEM_LIBCRYPTO=1 python3 -m pip install --no-binary :all: --verbose awscrt
+```
+
+( `--no-binary :all:` ensures you do not use the wheel from PyPI always uses AWS-LC)
+You can ignore all this on Windows and Apple platforms, where aws-crt-python
+uses the OS's default libraries for TLS and cryptography math.
```
graebm, line 68: paragraph breaks

```suggestion
( `--no-binary :all:` ensures you do not use the wheel from PyPI always uses AWS-LC)

You can ignore all this on Windows and Apple platforms, where aws-crt-python
uses the OS's default libraries for TLS and cryptography math.
```

.github/workflows/ci.yml
```diff
@@ -102,6 +102,18 @@ jobs:
           aws s3 cp s3://aws-crt-test-stuff/ci/${{ env.BUILDER_VERSION }}/linux-container-ci.sh ./linux-container-ci.sh && chmod a+x ./linux-container-ci.sh
           ./linux-container-ci.sh ${{ env.BUILDER_VERSION }} aws-crt-${{ env.LINUX_BASE_IMAGE }} build -p ${{ env.PACKAGE_NAME }} --compiler=${{ matrix.compiler }}
 
+  use-openssl:
```
graebm, line 105:
```suggestion
  use-system-libcrypto:
```

.github/workflows/ci.yml
```diff
@@ -102,6 +102,18 @@ jobs:
...
+    runs-on: ubuntu-20.04 # latest
+    steps:
+      - name: Build ${{ env.PACKAGE_NAME }}
+        env:
+          AWS_CRT_BUILD_USE_SYSTEM_LIBCRYPTO: '1'
+        run: |
+          python3 -c "from urllib.request import urlretrieve; urlretrieve('${{ env.BUILDER_HOST }}/${{ env.BUILDER_SOURCE }}/${{ env.BUILDER_VERSION }}/builder.pyz?run=${{ env.RUN }}', 'builder')"
+          chmod a+x builder
+          ./builder build -p ${{ env.PACKAGE_NAME }}
+
```
graebm, line 115:
```
# assert it's linked against the system's libcrypto.so
run: |
  AWSCRT_PATH=`python3 -c "import _awscrt; print(_awscrt.__file__)"`
  printf "AWSCRT_PATH: $AWSCRT_PATH\n"

  LINKED_AGAINST=`ldd $AWSCRT_PATH`
  printf "LINKED AGAINST:\n$LINKED_AGAINST\n"

  USES_LIBCRYPTO_SO=`echo "$LINKED_AGAINST" | grep 'libcrypto*.so' | head -1`
  test -n "$USES_LIBCRYPTO_SO"

```

.github/workflows/ci.yml
knightjoel: If memory serves, builder tries to clone the project repo before it does the installation of platform packages. This doesn't work on OpenBSD because git isn't available in the base OS install and has to be installed from packages.

.github/workflows/ci.yml
```diff
@@ -128,9 +140,6 @@ jobs:
   openbsd:
     runs-on: ubuntu-latest # macos would be faster. TODO: diagnose why some tests are failing
     steps:
-      - uses: actions/checkout@v3
-        with:
-          submodules: true
```
knightjoel, line 133: This change is causing the OpenBSD CI workflow to fail. OpenBSD doesn't ship git in the base install, however builder will try to do a git clone before it installs the platform packages. The workflow needs to check out the project in order to bootstrap builder.

README.md
```diff
@@ -12,25 +12,56 @@ Python 3 bindings for the AWS Common Runtime.
...
+### OpenSSL and LibCrypto (Unix only)
+
+aws-crt-python does not use OpenSSL for TLS.
+On Apple and Windows devices, the OS's default TLS library is used.
+On Unix devices, [s2n-tls](https://github.com/aws/s2n-tls) is used.
+But s2n-tls uses libcrypto, the cryptography math library bundled with OpenSSL.
+
+You can ignore all this on Windows and Apple platforms, where aws-crt-python
+uses the OS's default libraries for TLS and cryptography math.
+
```
graebm, line 46: Oops, this sentence is in here twice. Once here, and again at the end of the section. let's remove this one. I edit things to death, and sometimes leave in ghosts of old edits by accident
```suggestion
```

README.md
```diff
@@ -12,25 +12,56 @@ Python 3 bindings for the AWS Common Runtime.
...
+AWS-LC's code is included in the PyPI source package, 
+and the git repository includes it as a submodule.
+
+If you need aws-crt-python to use the libcrypto included on your system, 
+set environment variable `AWS_CRT_BUILD_USE_SYSTEM_LIBCRYPTO=1` while building from source:
+
+```sh
+AWS_CRT_BUILD_USE_SYSTEM_LIBCRYPTO=1 python3 -m pip install --no-binary :all: --verbose awscrt
+```
+( `--no-binary :all:` ensures you do not use the wheel from PyPI always uses AWS-LC)
```
graebm, line 60: another ghost of edits past
```suggestion
( `--no-binary :all:` ensures you do not use the precompiled wheel from PyPI)
```

setup.py
```diff
@@ -304,6 +312,8 @@ def awscrt_ext():
         # OpenBSD doesn't have librt; functions are found in libc instead.
         if not sys.platform.startswith('openbsd'):
             libraries += ['rt']
+        if using_system_libcrypto():
+            libraries += ['crypto']
```
graebm, line 316: utterly trivial: whitespace between the openbsd stuff and this
```suggestion
        # OpenBSD doesn't have librt; functions are found in libc instead.
        if not sys.platform.startswith('openbsd'):
            libraries += ['rt']

        if using_system_libcrypto():
            libraries += ['crypto']
```

graebm (approved):
fix & ship!

I'm going to make the `use-system-libcrypto` check required now...

PR #473, opened 2023-05-10
FreeBSD ci

TingDaoK:
- experiments about FreeBSD

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -200,8 +200,8 @@ jobs:
           envs: 'AWS_ACCESS_KEY_ID AWS_SECRET_ACCESS_KEY AWS_DEFAULT_REGION AWS_REGION'
           usesh: true
           run: |
-            pkg install python38 py38-pip py38-urllib3
-            python3 -c "from urllib.request import urlretrieve; urlretrieve('${{ env.BUILDER_HOST }}/${{ env.BUILDER_SOURCE }}/${{ env.BUILDER_VERSION }}/builder.pyz', 'builder')"
+            pkg install python39 py39-pip py39-urllib3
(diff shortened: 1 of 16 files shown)


.github/workflows/ci.yml
```diff
@@ -186,6 +186,28 @@ jobs:
...
+        with:
+          submodules: true
+      - name: Build ${{ env.PACKAGE_NAME }} + consumers
+        uses: vmactions/freebsd-vm@v0
+        with:
+          envs: 'AWS_ACCESS_KEY_ID AWS_SECRET_ACCESS_KEY AWS_DEFAULT_REGION AWS_REGION'
+          usesh: true
+          sync: rsync
+          copyback: false
+          run: |
```
DmitriyMusatkin, line 204: nit: this action has a prepare step, which might be a better place for installing stuff like pip

test/test_checksums.py
```diff
@@ -80,11 +81,13 @@ def test_crc32c_large_buffer(self):
         self.assertEqual(0xfb5b991d, val)
 
     def test_crc32c_huge_buffer(self):
-        # stress the internal logic that handles buffers larger than C's INT_MAX
+        if sys.platform.startswith('freebsd'):
```
DmitriyMusatkin, line 84: only fails for crc32c? do we want to skip other flex checksum cases for huge buffer?

DmitriyMusatkin (approved):

.github/workflows/ci.yml
```diff
@@ -7,8 +7,8 @@ on:
       - 'docs'
 
 env:
-  BUILDER_VERSION: v0.9.44
-  BUILDER_SOURCE: releases
+  BUILDER_VERSION: freebsd-ci
```
DmitriyMusatkin, line 10: dont forget to revert before pushing

test/test_checksums.py
TingDaoK: It only fails for crc32c for now, I skip the other huge_buffer test for safe.

PR #484, opened 2023-06-29
Update crash handler behavior to opt-in

waahm7:
*Description of changes:*
Our current crash handler unintentionally catches crashes from other Python libraries that utilize native code. The stacktrace dumped by the crash handler makes it look like our fault as it contains `PyInt_init` calls at the top of the stack trace. These calls, in reality, are simply function calls made to print the stack trace and printed as `PyInt_init` in release mode. To address this, this PR alters the crash handler to be an opt-in feature, disabling its automatic registration. Users can enable the crash handler by setting an env variable.


awscrt/common.py
```diff
@@ -18,3 +18,10 @@ def get_cpu_count_for_group(group_idx: int) -> int:
     Returns number of processors in a given group.
     """
     return _awscrt.get_cpu_count_for_group(group_idx)
+
+
+def install_crash_handler():
```
graebm, line 23: Seems easier to just query an environment variable from C
Vs this thing that customers need to call, which (means it can't catch initialization crashes, because initialization has already happend). And we need to worry about it being called multiple times. And how would a CLI user make this happen?

waahm7: Thanks, fixed.

source/module.c
```diff
@@ -804,6 +804,7 @@ static PyMethodDef s_module_methods[] = {
 
 static const char s_module_name[] = "_awscrt";
 PyDoc_STRVAR(s_module_doc, "C extension for binding AWS implementations of MQTT, HTTP, and friends");
+AWS_STATIC_STRING_FROM_LITERAL(s_crash_handler_env_var, "AWS_CRT_ENABLE_CRASH_HANDLER");
```
graebm, line 807: debatable: shorter name, similar to pre-existing "AWS_CRT_MEMORY_TRACING"?
```suggestion
AWS_STATIC_STRING_FROM_LITERAL(s_crash_handler_env_var, "AWS_CRT_CRASH_HANDLER");
```

source/module.c
```diff
@@ -828,11 +829,16 @@ PyMODINIT_FUNC PyInit__awscrt(void) {
     }
 
     s_init_allocator();
-    s_install_crash_handler();
 
     /* Don't report this memory when dumping possible leaks. */
     struct aws_allocator *nontracing_allocator = aws_default_allocator();
 
+    struct aws_string *crash_handler_env = NULL;
```
graebm, line 836: Speaking of "AWS_CRT_MEMORY_TRACING", we have [a hack](https://github.com/awslabs/aws-crt-python/blob/ce221859f37325d362c587ce2ef22b5ced3a4502/test/__init__.py#L8-L9) so that it's always turned on when tests run. let's do the same thing and always turn on the crash handler

README.md
```diff
@@ -66,3 +66,6 @@ Please note that on Mac, once a private key is used with a certificate, that cer
 ```
 static: certificate has an existing certificate-key pair that was previously imported into the Keychain. Using key from Keychain instead of the one provided.
 ```
+
+## Crash Handler
+You can enable the crash handler by setting the environment variable `AWS_CRT_ENABLE_CRASH_HANDLER=1`. This will generate crash dumps in the event of a fatal error.
```
graebm, line 71:
```suggestion
You can enable the crash handler by setting the environment variable `AWS_CRT_ENABLE_CRASH_HANDLER=1`. This will print the callstack to `stderr` in the event of a fatal error.
```

graebm (approved):

PR #498, opened 2023-08-15
Fix mqtt3 client shutdown callback

xiazhvera:
*Issue #, if available:*

*Description of changes:*


source/mqtt_client_connection.c
```diff
@@ -49,10 +49,6 @@ struct mqtt_connection_binding {
      * Lets us invoke callbacks on the python object without preventing the GC from cleaning it up. */
     PyObject *self_proxy;
 
-    /* To not run into a segfault calling on_close with the connection being freed before the callback
-     * can be invoked, we need to keep the PyCapsule alive. */
-    PyObject *self_capsule;
```
xiazhvera, line 54: `self_capsule` was a trick we used to keep the python object alive before the client tear down. As we now have the termination callback, we no longer need it.

source/mqtt_client_connection.c
```diff
@@ -374,9 +382,11 @@ PyObject *aws_py_mqtt_client_connection_new(PyObject *self, PyObject *args) {
     Py_DECREF(self_proxy);
 proxy_new_failed:
 use_websockets_failed:
+set_on_closed_failed:
```
bretambrose, line 385: I've come to hate this pattern for cleanup.  We did it early on, but it is far better (imo) to have a failure path that can safely be executed from all branch points.  If scoping is an issue, declarations can be moved up to the top of the function.

Would much rather see something like this:

```
on_failure:

   Py_XDECREF(self_proxy); 
   aws_mqtt_client_connection_release(py_connection->native);
   aws_mem_release(allocator, py_connection);
   return NULL;
```

source/mqtt_client_connection.c
```diff
@@ -345,40 +356,37 @@ PyObject *aws_py_mqtt_client_connection_new(PyObject *self, PyObject *args) {
...
-proxy_new_failed:
-use_websockets_failed:
-set_interruption_failed:
-set_connection_handlers_failed:
-    aws_mqtt_client_connection_release(py_connection->native);
-connection_new_failed:
-    aws_mem_release(allocator, py_connection);
+creation_failed:
+    Py_XDECREF(self_proxy);
+    if (py_connection) {
```
bretambrose, line 384: py_connection has to be non-null at this point.
Also, maybe just "on_error:" rather than "creation_failed:"

source/mqtt_client_connection.c
```diff
@@ -345,40 +356,37 @@ PyObject *aws_py_mqtt_client_connection_new(PyObject *self, PyObject *args) {
...
+        if (py_connection->native) {
```
bretambrose, line 385: You also don't need to null check this either.  release handles a null invocation.  The original suggestion would work verbatim.

bretambrose (approved):
Fix n ship

PR #507, opened 2023-10-07
Env Detection

JonathanHenson:
For p4d.24xlarge:
```
Is on Ec2 Instance nitro ? True
Instance Type = p4d.24xlarge
Is Optimized for System = True
```
For  trn1n.32xlarge:
```
Is on Ec2 Instance nitro ? True
Instance Type = trn1n.32xlarge
Is Optimized for System =  True
```
For p5.48xlarge:
```
Is on Ec2 Instance nitro ? True
Instance Type = p5.48xlarge
Is Optimized for System = True
```
For c5n.18xlarge:
```
Is on Ec2 Instance nitro ? True
Instance Type = c5n.18xlarge
Is Optimized for System = False
```
For t2.xlarge
```
Is on Ec2 Instance nitro ? {} False
Instance Type = {} None
Is Optimized for System = {} False
```


source/common.c
```diff
@@ -26,6 +29,110 @@ PyObject *aws_py_get_cpu_count_for_group(PyObject *self, PyObject *args) {
...
+    return NULL;
+}
+
+PyObject *aws_py_is_crt_s3_optimized_for_system(PyObject *self, PyObject *args) {
+    PyObject *env_capsule = NULL;
+    const char *instance_type_str = NULL;
+    Py_ssize_t instance_type_str_len = 0;
+
+    if (!PyArg_ParseTuple(args, "Oz#", &env_capsule, &instance_type_str, &instance_type_str_len)) {
+        return PyErr_AwsLastError();
```
graebm, line 111:
```suggestion
        return NULL;
```

awscrt/common.py
```diff
@@ -18,3 +18,23 @@ def get_cpu_count_for_group(group_idx: int) -> int:
...
+    def __init__(self):
+        self._env = _awscrt.load_system_environment()
+
+        if self.is_ec2_nitro_instance():
+            self._detected_instance_type = _awscrt.get_ec2_instance_type(self._env)
+        else:
+            self._detected_instance_type = None       
+
+    def is_ec2_nitro_instance(self) -> bool:    
+        return _awscrt.is_env_ec2(self._env)
```
graebm, line 33: Why mention "nitro" here, but not in the underlying C function?

I liked the `is_running_on_ec2` name better, FWIW

awscrt/common.py
```diff
@@ -18,3 +18,23 @@ def get_cpu_count_for_group(group_idx: int) -> int:
...
+    
+    def get_ec2_instance_type(self) -> str:        
+        return self._detected_instance_type
+    
+    def is_crt_s3_optimized_for_system_env(self) -> bool:
```
graebm, line 38: This is awkward:
`system_env.is_crt_s3_optimized_for_system_env()`

flip the words around to sound more natural:
`system_env.is_optimized_for_crt_s3()`

or it might feel more natural as a separate function in s3.py:
`awscrt.s3.is_optimized_for_system(system_env)`

awscrt/common.py
```diff
@@ -18,3 +18,23 @@ def get_cpu_count_for_group(group_idx: int) -> int:
...
+
+        if self.is_ec2_nitro_instance():
+            self._detected_instance_type = _awscrt.get_ec2_instance_type(self._env)
+        else:
+            self._detected_instance_type = None       
+
+    def is_ec2_nitro_instance(self) -> bool:    
+        return _awscrt.is_env_ec2(self._env)
+    
+    def get_ec2_instance_type(self) -> str:
```
graebm, line 35: the underylying C function is `get_virtualization_product_name()`

let's consistently call it "ec2" or "virtualization_product_name" but it's annoying to change terminology between the layers

awscrt/common.py
graebm: Or just have 1 public function: `awscrt.s3.is_optimized_for_system()`

and don't bother binding the aws_system_environment at all, just keep it as an implementation detail

we can have python cache the answer, so it only does work the first time it's called

awscrt/common.py
```diff
@@ -18,3 +18,23 @@ def get_cpu_count_for_group(group_idx: int) -> int:
     Returns number of processors in a given group.
     """
     return _awscrt.get_cpu_count_for_group(group_idx)
+
+class SystemEnvironment:
+
+    def __init__(self):
```
graebm, line 24: Inherit from NativeResource. we just do this to help tests track leaks
```suggestion
class SystemEnvironment(NativeResource):

    def __init__(self):
        super().__init__()
```

awscrt/common.py
JonathanHenson: ec2 and vritualization product name are not the same thing. ec2 is A vendor, and aws-c-common is mostly generic, I did move most of this to aws-c-s3 however and it makes more sense there. I can clean it up here.

awscrt/common.py
JonathanHenson: I'm just gonna pull out systemenvironment entirely and cache it internally as you suggested above.

awscrt/common.py
JonathanHenson: That's fine. I added verbosity here largely for aid in debugging when something goes wrong, we can at least ask the customer to add a python function to their script and it's much simpler to track down, than having to peel through logs or c source.

awscrt/common.py
JonathanHenson: It's really a nitro check, and ec2 is implied as a result. 

I'll make sure nitro is put in there and fuse it with the suggestion you made

source/common.c
```diff
@@ -26,6 +29,110 @@ PyObject *aws_py_get_cpu_count_for_group(PyObject *self, PyObject *args) {
...
+
+    struct aws_string *instance_type = aws_s3_get_ec2_instance_type(allocator, env);
+
+    if (instance_type) {
+        PyObject *ret_value = PyUnicode_FromAwsString(instance_type);
+        aws_string_destroy(instance_type);
+        return ret_value;
+    }
+
+    return NULL;
```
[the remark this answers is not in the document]
JonathanHenson: I don't think there's a complicated cardinality here. If we can't get the instance type we don't really care why. We can't use it and we go on. It's not an error to not be able to detect it, so I'm fine with NULL or the empty string. Empty string is probably a better experience though, so I prefer that.

source/common.c
```diff
@@ -7,6 +7,8 @@
 #include <aws/common/system_info.h>
 #include <aws/common/thread.h>
 
+const char *s_capsule_name_sys_env = "aws_system_environment";
+
```
graebm, line 11: trivial: not needed anymore
```suggestion
```

graebm (approved):

awscrt/s3.py
```diff
@@ -503,3 +505,39 @@ def create_default_s3_signing_config(*, region: str, credential_provider: AwsCre
         use_double_uri_encode=False,
         should_normalize_uri_path=False,
     )
+
+
+def is_running_on_ec2_nitro():
```
graebm: I don't see the point in this function

continuous-delivery/test-pip-install.py
```diff
@@ -1,7 +1,9 @@
...
 
-import awscrt.io
+import awscrt.s3
 
-print('if the next statement does not explode, the pip install was successful') 
-print('Is alpn supported? {}', awscrt.io.is_alpn_available())
+print("if the next statement does not explode, the pip install was successful")
+print(f"Is on Ec2 Instance nitro ? {awscrt.s3.is_running_on_ec2_nitro()}")
+print(f"Instance Type = {awscrt.s3.get_ec2_instance_type()}")
+print(f"Is Optimized for System = {awscrt.s3.is_optimized_for_system()}")
```
graebm: revert this
it makes it seem like our release pipeline somehow how depends on checking that it's running on EC2

graebm (approved):

PR #512, opened 2023-10-17
Add checksum settings for S3

graebm:

### DIFF
--- a/awscrt/s3.py
+++ b/awscrt/s3.py
@@ -12,7 +12,9 @@ from awscrt.http import HttpRequest
 from awscrt.io import ClientBootstrap, TlsConnectionOptions
 from awscrt.auth import AwsCredentialsProvider, AwsSignatureType, AwsSignedBodyHeaderType, AwsSignedBodyValue, AwsSigningAlgorithm, AwsSigningConfig
 import awscrt.exceptions
+from dataclasses import dataclass
 import threading
+from typing import Optional
 from enum import IntEnum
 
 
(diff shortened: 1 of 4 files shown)


awscrt/s3.py
```diff
@@ -130,6 +195,9 @@ def on_shutdown():
 
         s3_client_core = _S3ClientCore(bootstrap, credential_provider, signing_config, tls_connection_options)
 
+        if compute_content_md5 is None:
```
DmitriyMusatkin, line 198: do we really want to expose compute md5 on client level to clients? 
The way checksum is configured is kinda ugly on c side, so if we can avoid leaking that through binding, we should probably prefer that. 

ideally my thinking on this is that we should have per request config to set which checksum should be computed (either md5 or flex one) and where to put it (header or trailer). and then the logic is always the same, for single part, respect checksum they provided and compute one otherwise, and for multipart just always compute one

graebm: This mirrors how it's currently exposed in [aws-c-s3](https://github.com/awslabs/aws-c-s3/blob/f710806f290d9a4d34c4c5ce07be66d314b6fba2/include/aws/s3/s3_client.h#L355)

There's a TODO on how maybe it should live somewhere else, but currently this is the only way to control Content-MD5

source/s3_meta_request.c
```diff
@@ -346,20 +347,22 @@ PyObject *aws_py_s3_client_make_meta_request(PyObject *self, PyObject *args) {
...
+    int type;                                          /* i */
+    PyObject *signing_config_py;                       /* O */
+    PyObject *credential_provider_py;                  /* O */
+    const char *recv_filepath;                         /* z */
+    const char *send_filepath;                         /* z */
+    struct aws_byte_cursor region;                     /* s# */
+    enum aws_s3_checksum_algorithm checksum_algorithm; /* i */
+    enum aws_s3_checksum_location checksum_location;   /* i */
+    int validate_response_checksum;                    /* p - boolean predicate */
+    PyObject *py_core;                                 /* O */
```
TingDaoK, line 362: 💯 I like how this makes it much easier to check

awscrt/s3.py
TingDaoK: Ideally, we should move the md5 to the checksum config, however, as the other configurations doesn't apply to md5 (md5 only in header and we don't have a way to validate it form response). So, it seems like it may still be better to be its own thing

graebm: We talked and decided to remove this as an option. It would be better as a request-option, instead of a client-option. But that would be a lot of work, and flexible-checksums are better in every way. So let's make flexible-checksums the only thing we expose for now.

PR #519, opened 2023-11-06
Bind out cross-process lock with unit tests.

JonathanHenson:

### DIFF
--- a/awscrt/s3.py
+++ b/awscrt/s3.py
@@ -53,7 +53,14 @@ def crt_instance_lock_acquire(lock_scope_name):
     the lock is already held by another caller. Callers should use this value
     in a with block for automatic release when they're finished with it, or explicitly call
     release(). When the process exits, this lock will be released regardless of if 
-    release has been invoked. """
+    release has been invoked.
+    
+    Keyword Args:
+        lock_scope_name (str): Unique string identifying the caller holding the lock.
+        
+    Returns:
+        InstanceLock
+    """
     return InstanceLock(_awscrt.s3_instance_lock_acquire(lock_scope_name))
 
 
--- a/awscrt/s3.py
+++ b/awscrt/s3.py
@@ -18,6 +18,44 @@ import threading
 from typing import Optional
 from enum import IntEnum
 
+class InstanceLock(NativeResource):
+    """
+    Class representing an exclusive cross-process lock.
+    It is created by calling crt_instance_lock_acquire().
+
+    Recommended usage is to either explicitly call release() when the lock is no longer required,
+    or use this in a 'with' statement.
+
+    If the lock has not been explicitly released when the process exits, it will be released by 
+    the operating system.
+    """
+    def __init__(self, lock_handle):
+        super().__init__()
+        self._binding = lock_handle
+
+    def __enter__(self):
+        # do nothing as we already have the lock
+        return
+    
+    def release(self):
+        if self._binding != None:
+            _awscrt.s3_instance_lock_release(self._binding)
+            self._binding = None
+
+    def __exit__(self, exc_type, exc_value, exc_tb):
+        self.release()
+
+
+def crt_instance_lock_acquire(lock_scope_name):
+    """
+    Acquires an exclusive cross-process lock scoped by 'lock_scope_name'.
+    Throws a Runtime Error with error code: AWS_ERROR_MUTEX_CALLER_NOT_OWNER, if
+    the lock is already held by another caller. Callers should use this value
+    in a with block for automatic release when they're finished with it, or explicitly call
+    release(). When the process exits, this lock will be released regardless of if 
+    release has been invoked. """
+    return InstanceLock(_awscrt.s3_instance_lock_acquire(lock_scope_name))
+
 
 class S3RequestType(IntEnum):
     """The type of the AWS S3 request"""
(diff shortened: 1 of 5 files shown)


awscrt/s3.py
```diff
@@ -18,6 +18,44 @@
...
+
+
+def crt_instance_lock_acquire(lock_scope_name):
+    """
+    Acquires an exclusive cross-process lock scoped by 'lock_scope_name'.
+    Throws a Runtime Error with error code: AWS_ERROR_MUTEX_CALLER_NOT_OWNER, if
+    the lock is already held by another caller. Callers should use this value
+    in a with block for automatic release when they're finished with it, or explicitly call
+    release(). When the process exits, this lock will be released regardless of if 
+    release has been invoked. """
```
nateprewitt, line 56: Couple quick pieces of metadata to keep this inline with existing typing information.

```suggestion
    """
    Acquires an exclusive cross-process lock scoped by 'lock_scope_name'.
    Throws a Runtime Error with error code: AWS_ERROR_MUTEX_CALLER_NOT_OWNER, if
    the lock is already held by another caller. Callers should use this value
    in a with block for automatic release when they're finished with it, or explicitly call
    release(). When the process exits, this lock will be released regardless of if 
    release has been invoked.
    
    Keyword Args:
        lock_scope_name (str): Unique string identifying the caller holding the lock.
        
    Returns:
        InstanceLock
    """
```

test/test_s3.py
```diff
@@ -41,6 +45,53 @@
...
+        self.nonce = time.time()
+        super().setUp()
+
+    def test_with_statement(self):
+        nonce_str = 'lock_a_{}'.format(self.nonce)
+        with crt_instance_lock_acquire(nonce_str) as lock:
+            try:
+                new_lock = crt_instance_lock_acquire(nonce_str)
+                self.fail("Acquiring a lock by the same nonce should fail when it's already held")
+            except RuntimeError as e:
```
nateprewitt, line 69: Does the CRT have any formalized error hierarchy in Python? I haven't dug too deeply yet. My only concern here is RuntimeError is very generic so it will be difficult to do control flow handling with this without message introspection.

nateprewitt (commented):
Two minor notes, but otherwise looks good. I'll pull the branch down and try it out.

test/test_s3.py
JonathanHenson: @graebm  any thoughts here?

source/s3_client.c
```diff
@@ -37,6 +39,70 @@ PyObject *aws_py_s3_is_crt_s3_optimized_for_system(PyObject *self, PyObject *arg
...
+PyObject *aws_py_s3_instance_lock_release(PyObject *self, PyObject *args) {
+    struct aws_allocator *allocator = aws_py_get_allocator();
+
+    PyObject *lock_capsule; /* O */
+
+    if (!PyArg_ParseTuple(args, "O", &lock_capsule)) {
+        return NULL;
+    }
+
+    struct instance_lock_binding *lock_binding = PyCapsule_GetPointer(lock_capsule, s_capsule_name_s3_instance_lock);
```
graebm, line 96: trivial:
```suggestion
    struct instance_lock_binding *lock_binding = PyCapsule_GetPointer(lock_capsule, s_capsule_name_s3_instance_lock);
    if (!lock_binding) {
        return NULL;
    }
```

source/s3_client.c
```diff
@@ -37,6 +39,70 @@ PyObject *aws_py_s3_is_crt_s3_optimized_for_system(PyObject *self, PyObject *arg
...
+    if (!lock) {
+        return PyErr_AwsLastError();
+    }
+
+    struct instance_lock_binding *binding = aws_mem_calloc(allocator, 1, sizeof(struct instance_lock_binding));
+    binding->lock = lock;
+
+    PyObject *capsule = PyCapsule_New(binding, s_capsule_name_s3_instance_lock, s_s3_instance_lock_destructor);
+    if (!capsule) {
+        aws_mem_release(allocator, binding);
```
graebm, line 80:
```suggestion
    if (!capsule) {
        aws_cross_process_lock_release(lock);
        aws_mem_release(allocator, binding);
```

test/test_s3.py
graebm: It's not great. It's the same problem we have in every language binding. We didn't code-gen error classes, so catching specific CRT errors means users must catch a RuntimeError and introspect it.

In this specific case, we could dance around the issue and just return None or False if the C function returns NULL, since we do often expect it to fail 🤷‍♀️

awscrt/s3.py
```diff
@@ -18,6 +18,44 @@
 from typing import Optional
 from enum import IntEnum
 
+class InstanceLock(NativeResource):
```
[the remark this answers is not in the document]
JonathanHenson: This is more idiomatic. I can do that.

test/test_s3.py
graebm: I'm not happy about this. I really want to do a mildly-breaking-change to all our bindings and address this issue. But it's non-trivial :|

awscrt/s3.py
```diff
@@ -18,6 +18,51 @@
...
```
graebm, line 21: Name it consistently with the C code
```suggestion
class CrossProcessLock(NativeResource):
```

awscrt/s3.py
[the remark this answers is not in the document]
JonathanHenson: done.

graebm (approved):

PR #520, opened 2023-11-08
Added function for returning the recommended max bandwidth for the cr…

JonathanHenson:
…t s3 client based on detected environment.

### DIFF
--- a/awscrt/s3.py
+++ b/awscrt/s3.py
@@ -178,7 +178,7 @@ class S3Client(NativeResource):
 
         throughput_target_gbps (Optional[float]): Throughput target in
             Gigabits per second (Gbps) that we are trying to reach.
-            You can also use `get_recommended_throughput_target_gbps` to get recommended value for your system.
+            You can also use `get_recommended_throughput_target_gbps()` to get recommended value for your system.
             10.0 Gbps by default (may change in future)
     """
 
--- a/awscrt/s3.py
+++ b/awscrt/s3.py
@@ -597,7 +597,7 @@ def get_recommended_max_throughput_gbps():
     # Currently the CRT returns 0 if it was unable to make a good guess on configuration. Pre-known configs, have this value set.
     # Eventually, the CRT will make a full calculation based on NIC and CPU configuration, but until then handle 0.
     max_value = _awscrt.s3_get_recommended_max_throughput_gbps()
-    if abs(max_value) > 0:
+    if max_value > 0:
         return max_value
     else:
         return 5.0
--- a/awscrt/s3.py
+++ b/awscrt/s3.py
@@ -178,7 +178,8 @@ class S3Client(NativeResource):
 
         throughput_target_gbps (Optional[float]): Throughput target in
             Gigabits per second (Gbps) that we are trying to reach.
-            (10.0 Gbps by default)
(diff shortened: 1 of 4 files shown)


waahm7 (approved):

awscrt/s3.py
```diff
@@ -586,3 +586,18 @@ def is_optimized_for_system():
             for the current system.
     """
     return _awscrt.s3_is_crt_s3_optimized_for_system()
+
+
+def get_recommended_max_throughput_gbps():
+    """
+        Returns:
+        Recommended max throughput in Gbps for the CRT S3 client based on detected system configuration as a PyFloat.
+        If the best throughput configuration is unknown, returns 5Gbps.
```
graebm, line 595: Spell out gigabits per second in the docs. I've seen plenty of confusion in various codebases on what "Gb" means.

Also, PyFloat is a c-api term. It's just called `float` in python. And you could just omit this if you put type annotation on the function `-> float:` and/or said the default was `5.0`
```suggestion
    Returns:
        Recommended throughput, in gigabits per second, based on detected system configuration.
        If the best throughput configuration is unknown, returns 5.0 Gbps.
        Use this as the S3Client's `throughput_target_gbps`.
``````

File: awscrt/s3.py [hunk 2]
```diff
@@ -586,3 +586,18 @@ def is_optimized_for_system():
...
+    """
+    # Currently the CRT returns 0 if it was unable to make a good guess on configuration. Pre-known configs, have this value set.
+    # Eventually, the CRT will make a full calculation based on NIC and CPU configuration, but until then handle 0.
+    max_value = _awscrt.s3_get_recommended_max_throughput_gbps()
+    if abs(max_value) > 0:
+        return max_value
+    else:
+        return 5.0
```
graebm, line 603: highly debatable: should we use this new functionality in the S3Client constructor if the user doesn't pass throughput_target_gbps?

or should we use this down in C, instead of the current behavior of defaulting to 10?

graebm: if not, let's at least add to the S3Client constructor's docs, to mention this function

waahm7: Discussed offline, I have updated it to return None so that we can keep the default value behavior in one place

awscrt/s3.py
```diff
@@ -586,3 +586,18 @@ def is_optimized_for_system():
             for the current system.
     """
     return _awscrt.s3_is_crt_s3_optimized_for_system()
+
+
+def get_recommended_max_throughput_gbps():
```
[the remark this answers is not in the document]
waahm7: I have updated it to `get_recommended_throughput_target_gbps`, as this value might differ from the system's throughput.

awscrt/s3.py
```diff
@@ -178,7 +178,8 @@ class S3Client(NativeResource):
 
         throughput_target_gbps (Optional[float]): Throughput target in
             Gigabits per second (Gbps) that we are trying to reach.
-            (10.0 Gbps by default)
+            You can also use `get_recommended_throughput_target_gbps` to get recommended value for your system.
```
graebm, line 181:
```suggestion
            You can also use `get_recommended_throughput_target_gbps()` to get recommended value for your system.
```

graebm (approved):

PR #521, opened 2023-11-09
API to get list of pre optimized systems

waahm7:
*Description of changes:*
Binds the `aws_s3_get_platforms_with_recommended_config()`

### DIFF
--- a/awscrt/s3.py
+++ b/awscrt/s3.py
@@ -586,3 +586,12 @@ def is_optimized_for_system():
             for the current system.
     """
     return _awscrt.s3_is_crt_s3_optimized_for_system()
+
+
+def get_platforms_with_recommended_config():
+    """
+        Returns a list of platforms identifiers such as EC2 instance types that have a recommended throughput target in gigabits per second (Gbps).
+        Returns:
+            list[str]: A list of platform identifiers with recommended throughput configurations.
+    """
+    return _awscrt.s3_get_platforms_with_recommended_config()
--- a/awscrt/s3.py
+++ b/awscrt/s3.py
@@ -614,4 +614,3 @@ def get_recommended_throughput_target_gbps() -> Optional[float]:
         return max_value
     else:
         return None
-
--- a/awscrt/s3.py
+++ b/awscrt/s3.py
@@ -589,6 +589,16 @@ def is_optimized_for_system():
     return _awscrt.s3_is_crt_s3_optimized_for_system()
 
 
+def get_optimized_platforms():
+    """
+    Returns:
+        A list[str] of platform identifiers, such as EC2 instance types, for which S3 client is pre-optimized
+        and have a recommended throughput_target_gbps. You can use `get_recommended_throughput_target_gbps()`
+        to obtain the recommended throughput_target_gbps for those platforms.
+    """
+    return _awscrt.s3_get_optimized_platforms()
+
+
 def get_recommended_throughput_target_gbps() -> Optional[float]:
     """
     Returns:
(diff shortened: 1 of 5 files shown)


awscrt/s3.py
```diff
@@ -589,15 +589,26 @@ def is_optimized_for_system():
     return _awscrt.s3_is_crt_s3_optimized_for_system()
 
 
+def get_optimized_platforms():
```
waahm7, line 592: I also considered the following names:
- s3_get_platforms_with_recommended_config (similar to C but no concept of config in Pythong)
- get_platforms_with_recommended_throughput_target() (Too verbose and we might add more things in future)

graebm: Given that the other public function is named `is_optimized_for_system()`, this is fine

if we were going to name this differently, then the other should be named differently too

source/s3_client.c
```diff
@@ -48,6 +48,31 @@ PyObject *aws_py_s3_get_recommended_throughput_target_gbps(PyObject *self, PyObj
...
+
+    PyObject *py_list = PyList_New(list_length);
+    AWS_FATAL_ASSERT(py_list && "platform list allocation failed");
+
+    for (size_t i = 0; i < list_length; ++i) {
+        struct aws_byte_cursor cursor;
+        if (aws_array_list_get_at(&platform_list, &cursor, i) == AWS_OP_SUCCESS) {
+            PyObject *platform_str = PyUnicode_FromStringAndSize((char *)cursor.ptr, cursor.len);
+            AWS_FATAL_ASSERT(platform_str && "platform str allocation failed");
+            PyList_SetItem(py_list, i, platform_str); /* Steals a Reference */
```
graebm, line 67: You must actually handle errors from Py functions. They can fail during interpreter shutdown.

waahm7: @graebm Which function are you talking about? I looked at https://github.com/awslabs/aws-crt-python/blob/main/source/websocket.c#L205 for fatal asserts and noticed that we don't handle the return value of PyList_SetItem in most places.

source/s3_client.c
```diff
@@ -48,6 +48,46 @@ PyObject *aws_py_s3_get_recommended_throughput_target_gbps(PyObject *self, PyObj
...
+    (void)args;
+
+    bool success = false;
+    struct aws_array_list platform_list = aws_s3_get_platforms_with_recommended_config();
+
+    size_t list_length = aws_array_list_length(&platform_list);
+
+    PyObject *py_list = PyList_New(list_length);
+    if (!py_list) {
+        PyErr_SetString(PyExc_MemoryError, "Failed to allocate memory for platform list");
```
graebm, line 62: you don't need to set an exception. PyXYZ() functions will have already set one if they fail
```suggestion
```

source/s3_client.c
```diff
@@ -48,6 +48,46 @@ PyObject *aws_py_s3_get_recommended_throughput_target_gbps(PyObject *self, PyObj
...
+        goto clean_up;
+    }
+
+    for (size_t i = 0; i < list_length; ++i) {
+        struct aws_byte_cursor cursor;
+        if (aws_array_list_get_at(&platform_list, &cursor, i) == AWS_OP_SUCCESS) {
+            PyObject *platform_str = PyUnicode_FromAwsByteCursor(&cursor);
+            if (!platform_str) {
+                Py_DECREF(py_list);
+                PyErr_SetString(PyExc_MemoryError, "Failed to convert platform string to PyObject");
```
graebm, line 72: likewise, an exception is already set
```suggestion
```

source/s3_client.c
```diff
@@ -48,6 +48,46 @@ PyObject *aws_py_s3_get_recommended_throughput_target_gbps(PyObject *self, PyObj
...
+                goto clean_up;
+            }
+            if (PyList_SetItem(py_list, i, platform_str) /* Steals a Reference */) {
+                /* PyList_SetItem failed, platform_str reference not stolen so DECREF it */
+                Py_DECREF(platform_str);
+                Py_DECREF(py_list);
+                PyErr_SetString(PyExc_RuntimeError, "Failed to set item in the plaform list");
+                goto clean_up;
+            }
```
graebm, line 81: you can assume this one succeeds, it can only fail due to indexing errors
```suggestion
            PyList_SetItem(py_list, i, platform_str); /* Steals a Reference */
```

graebm (approved):

PR #522, opened 2023-11-10
S3ResponseError

graebm:
**Issue:**
The exceptions delivered from the S3Client were lacking information like the HTTP response code, headers, & body. You could get the extra information from the `on_done` callback, but it would be nice to get it from the exception, so you could use the `request.finished_future` instead of the `on_done` callback if you wanted.

**Description of changes:**
Add new error type: `awscrt.s3.S3ResponseError`, which inherits from `awscrt.exceptions.AwsCrtError`. If we have information about the failed request, then use an `S3ResponseError` instead.

Debatable:
- Should this live in `awscrt.exceptions` or `awscrt.s3`?

### DIFF
--- a/awscrt/s3.py
+++ b/awscrt/s3.py
@@ -464,6 +464,7 @@ class S3ResponseError(awscrt.exceptions.AwsCrtError):
         status_code (int): HTTP response status code.
         headers (list[tuple[str, str]]): Headers from HTTP response.
         body (Optional[bytes]): Body of HTTP response (if any).
+            This is usually XML. It may be None in the case of a HEAD response.
         code (int): CRT error code.
         name (str): CRT error name.
         message (str): CRT error message.
@@ -546,7 +547,8 @@ class _S3RequestCore:
         if error_code:
             error = awscrt.exceptions.from_code(error_code)
 
-            # Make this into an S3ResponseError, if possible
+            # If the failure was due to a response, make it into an S3ResponseError.
+            # When failure is due to a response, its headers are always included.
             if isinstance(error, awscrt.exceptions.AwsCrtError) \
                     and status_code is not None \
                     and error_headers is not None:
--- a/awscrt/s3.py
+++ b/awscrt/s3.py
@@ -15,7 +15,7 @@ from awscrt.auth import AwsCredentialsProvider, AwsSignatureType, AwsSignedBodyH
 import awscrt.exceptions
 from dataclasses import dataclass
 import threading
-from typing import Optional
+from typing import List, Optional, Tuple
 from enum import IntEnum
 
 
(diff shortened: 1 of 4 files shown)


awscrt/s3.py
```diff
@@ -517,13 +545,18 @@ def _on_finish(self, error_code, status_code, error_headers, error_body):
...
-                try:
-                    extra_message = ". Body from error request is: " + str(error_body)
-                    error.message = error.message + extra_message
-                except BaseException:
-                    pass
+
+            # Make this into an S3ResponseError, if possible
+            if isinstance(error, awscrt.exceptions.AwsCrtError) \
+                    and status_code is not None \
+                    and error_headers is not None:
```
waahm7, line 552: Do we know if there won't be any case where status_code is `AWS_ERROR_S3_INVALID_RESPONSE_STATUS` but error_headers is None and error_body is present? Why assert on `error_headers` at all? I think we should just check if the status_code is `AWS_ERROR_S3_INVALID_RESPONSE_STATUS` and return whatever S3 gave us.

awscrt/s3.py
```diff
@@ -454,6 +454,34 @@ def cancel(self):
...
+class S3ResponseError(awscrt.exceptions.AwsCrtError):
+    '''
+    An error response from S3.
+
+    Subclasses :class:`awscrt.exceptions.AwsCrtError`.
+
+    Attributes:
+        status_code (int): HTTP response status code.
+        headers (list[tuple[str, str]]): Headers from HTTP response.
+        body (Optional[bytes]): Body of HTTP response (if any).
```
waahm7, line 466: trivial: We can add more details indicating it is in XML, and that it can be empty/null for cases like HeadRequest where S3 doesn't give us anything.

awscrt/s3.py
```diff
@@ -517,13 +545,18 @@ def _on_finish(self, error_code, status_code, error_headers, error_body):
...
+                    and status_code is not None \
+                    and error_headers is not None:
+                error = S3ResponseError(
+                    code=error.code,
+                    name=error.name,
+                    message=error.message,
+                    status_code=status_code,
+                    headers=error_headers,
+                    body=error_body)
             self._finished_future.set_exception(error)
```
waahm7, line 560: How about always throwing an `S3ResponseError`, which will contain as much information as we have, instead of two different exceptions?

waahm7: This is also helpful for cases when we add cause to `S3ResponseError`

awscrt/s3.py
graebm: I'll add better comment explaining my logic

graebm: There are several aws-error-codes besides AWS_ERROR_S3_INVALID_RESPONSE_STATUS where the failure is due to a response

awscrt/s3.py
graebm: The idea is to only raise S3ResponseError with the extra data when we actually have the extra data

waahm7 (approved):

PR #529, opened 2023-11-21
Bind out mem limiter

DmitriyMusatkin:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/awscrt/s3.py
+++ b/awscrt/s3.py
@@ -182,7 +182,7 @@ class S3Client(NativeResource):
             10.0 Gbps by default (may change in future)
 
         memory_limit (Optional[int]): Memory limit, in bytes, of how much memory
-            client can use for buffering data for requests. 
+            client can use for buffering data for requests.
             Default values scale with target throughput and are currently
             between 2GBs and 8GBs (might change in future.)
     """
--- a/awscrt/s3.py
+++ b/awscrt/s3.py
@@ -184,7 +184,7 @@ class S3Client(NativeResource):
         memory_limit (Optional[int]): Memory limit, in bytes, of how much memory
             client can use for buffering data for requests.
             Default values scale with target throughput and are currently
-            between 2GBs and 8GBs (may change in future)
+            between 2GiB and 8GiB (may change in future)
     """
 
     __slots__ = ('shutdown_event', '_region')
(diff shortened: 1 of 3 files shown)


TingDaoK (approved):

awscrt/s3.py
```diff
@@ -180,6 +180,11 @@ class S3Client(NativeResource):
             Gigabits per second (Gbps) that we are trying to reach.
             You can also use `get_recommended_throughput_target_gbps()` to get recommended value for your system.
             10.0 Gbps by default (may change in future)
+
+        memory_limit (Optional[int]): Memory limit, in bytes, of how much memory
+            client can use for buffering data for requests.
+            Default values scale with target throughput and are currently
+            between 2GBs and 8GBs (might change in future.)
```
waahm7, line 187: Should we have a minimum value? Like anything less than 2GB will be bumped up to 2GB.

waahm7 (approved):

awscrt/s3.py
waahm7, line 187: trivial: consistent with `throughput_target_gbps` docs wording.
```suggestion
            between 2GBs and 8GBs (may change in future)
```

DmitriyMusatkin: I though about it before, but im on a fence on that. Mostly because i didnt want to complicate logic. Right now its fairly straight forward, and we figure out number if you dont provide one and we respect what you provided otherwise (with some sanity checking). Lets backlog it for now?

awscrt/s3.py
```diff
@@ -180,6 +180,11 @@ class S3Client(NativeResource):
             Gigabits per second (Gbps) that we are trying to reach.
             You can also use `get_recommended_throughput_target_gbps()` to get recommended value for your system.
             10.0 Gbps by default (may change in future)
+
+        memory_limit (Optional[int]): Memory limit, in bytes, of how much memory
+            client can use for buffering data for requests.
+            Default values scale with target throughput and are currently
+            between 2GBs and 8GBs (may change in future)
```
graebm, line 187:
```suggestion
            between 2GiB and 8GiB (may change in future)
```
or if that seems pedantic, at least remove the "s"
```suggestion
            between 2GB and 8GB (may change in future)
```

graebm (approved):

PR #540, opened 2023-12-29
Shared Subscription tests

alfred2g:
Shared Subscription tests


test/test_mqtt5.py
```diff
@@ -171,7 +172,6 @@ def test_client_creation_maximum(self):
             receive_maximum=1000,
             maximum_packet_size=10000,
             will_delay_interval_sec=1000,
-            will=publish_packet,
```
sbSteveK, line 174: Why was this removed? This test is attempting a creation with everything that can possibly be set including the will.

sbSteveK (approved):
add will back to test_client_creation_maximum and ship

PR #554, opened 2024-04-05
Latest Submodules

waahm7:


.python-version
```diff
@@ -0,0 +1 @@
+3.10.11
```
TingDaoK, line 1: I guess this is why CI failing. we should ignore those files from IDE. we could add it to the .gitignore.

waahm7: Oops, my bad.

TingDaoK (approved):

PR #559, opened 2024-04-22
CBOR bindings

TingDaoK:
*Issue #, if available:*

*Description of changes:*


source/cbor.c
```diff
@@ -0,0 +1,774 @@
...
+S_ENCODER_WRITE_PYOBJECT(struct aws_byte_cursor, aws_byte_cursor_from_pyunicode, text)
+S_ENCODER_WRITE_PYOBJECT(bool, PyObject_IsTrue, bool)
+
+S_ENCODER_WRITE_PYOBJECT(uint64_t, PyLong_AsUnsignedLongLong, array_start)
+S_ENCODER_WRITE_PYOBJECT(uint64_t, PyLong_AsUnsignedLongLong, map_start)
+S_ENCODER_WRITE_PYOBJECT(uint64_t, PyLong_AsUnsignedLongLong, tag)
+
+PyObject *aws_py_cbor_encoder_write_unsigned_int(PyObject *self, PyObject *args) {
+    PyObject *pylong;
+    S_ENCODER_METHOD_START("O", &pylong);
```
DmitriyMusatkin, line 111: the only thing that changes is the last part of s_cbor_encoder_write_pyobject_as? maybe more more of that into macro and let macro just call correct s_cbor_encoder_write_pyobject_as?

having a macro thats just a partial method kinda scares me

source/cbor.c
```diff
@@ -0,0 +1,774 @@
...
+    Py_RETURN_NONE;
+}
+
+static PyObject *s_cbor_encoder_write_pylist(struct encoder_binding *encoder_binding, PyObject *py_list) {
+    Py_ssize_t size = PyList_Size(py_list);
+    aws_cbor_encoder_write_array_start(encoder_binding->native, (size_t)size);
+    for (Py_ssize_t i = 0; i < size; i++) {
+        PyObject *item = PyList_GetItem(py_list, i);
+        if (!item) {
+            return NULL;
```
DmitriyMusatkin, line 204: so we leave it in broken state if element is not valid?

source/cbor.c
```diff
@@ -0,0 +1,774 @@
...
+    PyObject *py_capsule = PyCapsule_New(decoder, s_capsule_name_cbor_decoder, s_cbor_decoder_capsule_destructor);
+    if (!py_capsule) {
+        aws_cbor_decoder_destroy(decoder);
+        return NULL;
+    }
+
+    return py_capsule;
+}
+
+#define S_GET_DECODER()                                                                                                \
```
DmitriyMusatkin, line 333: nit: but im not a big fan of macros inserting random code into functions, can we make macros more like a generic function that returns a value rathere then relying on side effects

source/cbor.c
```diff
@@ -0,0 +1,774 @@
...
+/**
+ * Generic helper to convert a cbor encoded data to PyObject
+ */
+static PyObject *s_cbor_decoder_pop_next_tag_to_pyobject(struct aws_cbor_decoder *decoder) {
+    uint64_t out_tag_val = 0;
+    if (aws_cbor_decoder_pop_next_tag_val(decoder, &out_tag_val)) {
+        return PyErr_AwsLastError();
+    }
+    /* TODO: implement those tags */
+    switch (out_tag_val) {
```
DmitriyMusatkin, line 641: so we currently support no tags? should we be more explcit in doc on what function supports?

DmitriyMusatkin (approved):

awscrt/cbor.py
```diff
@@ -0,0 +1,226 @@
...
+        """Write an int as cbor formatted,
+            val less than -2^64 will be encoded as Negative bignum for CBOR
+            val between -2^64 to -1, inclusive, will be encode as negative integer for CBOR
+            val between 0 to 2^64 - 1, inclusive, will be encoded as unsigned integer for CBOR
+            val greater than 2^64 - 1 will be encoded as Unsigned bignum for CBOR
+
+        Args:
+            val (int): value to be encoded and written to the encoded data.
+        """
+        assert isinstance(val, int)
```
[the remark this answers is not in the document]
TingDaoK: The function signature is just a type hint, it won't enforce anything. Just for documentation and code clarity. 

https://docs.python.org/3/library/typing.html

>Note
The Python runtime does not enforce function and variable type annotations. They can be used by third party tools such as [type checkers](https://docs.python.org/3/glossary.html#term-static-type-checker), IDEs, linters, etc.

awscrt/cbor.py
```diff
@@ -0,0 +1,226 @@
...
+        """Add a start of array element.
+            A legistic with the `number_entries`
+            for the cbor data items to be included in the array.
+            `number_entries` should 0 to 2^64 inclusive.
+            Otherwise, overflow will be raised.
+
+        Args:
+            number_entries (int): number of entries in the array to be written
+        """
+        if number_entries < 0 or number_entries > 2**64:
```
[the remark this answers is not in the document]
TingDaoK: The implementation will not allocate the whole array, it appends to the buffer when user keeps writing to it. So, the size of the array won't cause any allocation problem.

test/test_cbor.py
```diff
@@ -0,0 +1,263 @@
...
+
+        decoder = AwsCborDecoder(encoder.get_encoded_data())
+
+        # Temp val only for easier to debug.
+        t = decoder.pop_next_data_item()
+        self.assertEqual(val_to_write, t)
+
+    def test_cbor_encode_decode_indef(self):
+        encoder = AwsCborEncoder()
+        numerics = [-100.12, 100.0, -100, 100, 2**64 - 1, -2**64, 18446744073709551616.0]
```
DmitriyMusatkin, line 74: why are all those cases in a separate test from the json based ones?

DmitriyMusatkin (approved):

PR #560, opened 2024-04-25
Treat correlation data as binary not a nullable string

bretambrose:
Non utf-8 correlation data was leading to utf-8 decode errors when calling from the binding to _on_publish in the MQTT5 client.

* Deprecate correlation_data in favor of correlation_data_bytes
* Test various combinations of new/deprecated values to ensure backwards compatible behavior

### DIFF
--- a/awscrt/mqtt5.py
+++ b/awscrt/mqtt5.py
@@ -1104,8 +1104,8 @@ class PublishPacket:
         message_expiry_interval_sec (int): Sent publishes - indicates the maximum amount of time allowed to elapse for message delivery before the server should instead delete the message (relative to a recipient). Received publishes - indicates the remaining amount of time (from the server's perspective) before the message would have been deleted relative to the subscribing client. If left None, indicates no expiration timeout.
         topic_alias (int): An integer value that is used to identify the Topic instead of using the Topic Name.  On outbound publishes, this will only be used if the outbound topic aliasing behavior has been set to Manual.
         response_topic (str): Opaque topic string intended to assist with request/response implementations.  Not internally meaningful to MQTT5 or this client.
-        correlation_data (Any): Opaque binary data used to correlate between publish messages, as a potential method for request-response implementation.  Not internally meaningful to MQTT5.  For outbound publishes, this can be either a string or binary data.  For incoming publishes, this will be a utf8 string (if correlation data exists and it's convertible to utf-8) or None (either it didn't exist, or did but wasn't convertible)
-        correlation_data_bytes (Any): Opaque binary data used to correlate between publish messages, as a potential method for request-response implementation.  Not internally meaningful to MQTT5.  For outbound publishes, this field is ignored.  For incoming publishes, this will be binary data if correlation data is set, otherwise it will be None.
+        correlation_data (Any): Deprecated, use `correlation_data_bytes` instead.  Opaque binary data used to correlate between publish messages, as a potential method for request-response implementation.  Not internally meaningful to MQTT5.  For incoming publishes, this will be a utf8 string (if correlation data exists and it's convertible to utf-8) or None (either it didn't exist, or did but wasn't convertible)
+        correlation_data_bytes (Optional[bytes]): Opaque binary data used to correlate between publish messages, as a potential method for request-response implementation.  Not internally meaningful to MQTT5.  For outbound publishes, this field takes priority over `correlation_data`.  For incoming publishes, this will be binary data if correlation data is set, otherwise it will be None.
         subscription_identifiers (Sequence[int]): The subscription identifiers of all the subscriptions this message matched.
         content_type (str): Property specifying the content type of the payload.  Not internally meaningful to MQTT5.
         user_properties (Sequence[UserProperty]): List of MQTT5 user properties included with the packet.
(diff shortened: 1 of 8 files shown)


awscrt/mqtt5.py
```diff
@@ -1117,7 +1118,8 @@ class PublishPacket:
     message_expiry_interval_sec: int = None
     topic_alias: int = None
     response_topic: str = None
-    correlation_data: Any = None   # Unicode objects are converted to C strings using 'utf-8' encoding
+    correlation_data: Any = None   # Outgoing publishes: may be a string or binary data.  Incoming publishes: a string if correlation data exists on the packet and is convertible to utf-8
+    correlation_data_bytes: Any = None  # Incoming publishes only: binary data if correlation data exists on the packet
```
sbSteveK, line 1122: Could we set this to `Optional[bytes]` instead of `Any` ? It looks like we intend for this to only ever be [bytes] or None.

awscrt/mqtt5.py
```diff
@@ -1104,7 +1104,8 @@ class PublishPacket:
...
         response_topic (str): Opaque topic string intended to assist with request/response implementations.  Not internally meaningful to MQTT5 or this client.
-        correlation_data (Any): Opaque binary data used to correlate between publish messages, as a potential method for request-response implementation.  Not internally meaningful to MQTT5.
+        correlation_data (Any): Opaque binary data used to correlate between publish messages, as a potential method for request-response implementation.  Not internally meaningful to MQTT5.  For outbound publishes, this can be either a string or binary data.  For incoming publishes, this will be a utf8 string (if correlation data exists and it's convertible to utf-8) or None (either it didn't exist, or did but wasn't convertible)
+        correlation_data_bytes (Any): Opaque binary data used to correlate between publish messages, as a potential method for request-response implementation.  Not internally meaningful to MQTT5.  For outbound publishes, this field is ignored.  For incoming publishes, this will be binary data if correlation data is set, otherwise it will be None.
```
graebm, line 1108: `correlation_data_bytes` should be `Optional[bytes]`, right?

awscrt/mqtt5.py
```diff
@@ -1104,7 +1104,8 @@ class PublishPacket:
...
         topic_alias (int): An integer value that is used to identify the Topic instead of using the Topic Name.  On outbound publishes, this will only be used if the outbound topic aliasing behavior has been set to Manual.
         response_topic (str): Opaque topic string intended to assist with request/response implementations.  Not internally meaningful to MQTT5 or this client.
-        correlation_data (Any): Opaque binary data used to correlate between publish messages, as a potential method for request-response implementation.  Not internally meaningful to MQTT5.
+        correlation_data (Any): Opaque binary data used to correlate between publish messages, as a potential method for request-response implementation.  Not internally meaningful to MQTT5.  For outbound publishes, this can be either a string or binary data.  For incoming publishes, this will be a utf8 string (if correlation data exists and it's convertible to utf-8) or None (either it didn't exist, or did but wasn't convertible)
```
graebm, line 1107: debatable: mark as deprecated

awscrt/mqtt5.py
graebm, line 1: Still need to update [mqtt5.Client.publish(...)](https://github.com/awslabs/aws-crt-python/blob/9e04b94ad9459b23d0bd5a49f78d1ea17b65d7f3/awscrt/mqtt5.py#L1873-L1874) to use correlation_data_bytes

graebm: Also, in [Client init](https://github.com/awslabs/aws-crt-python/blob/9e04b94ad9459b23d0bd5a49f78d1ea17b65d7f3/awscrt/mqtt5.py#L1776-L1777) need to use `will.correlation_data_bytes`

graebm: Maybe add some helper function to the PublishPacket that will return you bytes from whichever one is set? and use that when passing data down to C

bretambrose: Not using the bytes variant is intentional

awscrt/mqtt5.py
bretambrose: Well, it's not deprecated though

awscrt/mqtt5.py
bretambrose: The native unpacking also allows strings.  Is there a way to express `Optional[bytes | string]` ?

awscrt/mqtt5.py
```diff
@@ -1861,6 +1875,7 @@ def puback(error_code, qos, reason_code, reason_string, user_properties_tuples):
                                      publish_packet.message_expiry_interval_sec,
                                      publish_packet.topic_alias,
                                      publish_packet.response_topic,
+                                     publish_packet.correlation_data_bytes,
                                      publish_packet.correlation_data,
```
graebm, line 1879: debatable: possibly simpler to just pass one or the other, like:
```suggestion
                                     publish_packet.correlation_data_bytes or publish_packet.correlation_data,
```

and on the C-side, just take 1 arg like:
 z* ([str](https://docs.python.org/3/library/stdtypes.html#str), [bytes-like object](https://docs.python.org/3/glossary.html#term-bytes-like-object) or None) [Py_buffer]

graebm (approved):

awscrt/mqtt5.py
bretambrose: That would simplify the C side significantly

bretambrose: Done

PR #576, opened 2024-07-08
Support machines with multiple NICs

waahm7:
*Description of changes:*
Adds an option to take a list of interface names and distribute connections across multiple NIC.

### DIFF
--- a/awscrt/s3.py
+++ b/awscrt/s3.py
@@ -206,8 +206,12 @@ class S3Client(NativeResource):
             Default values scale with target throughput and are currently
             between 2GiB and 8GiB (may change in future)
 
-        network_interface_names: (Optional[list(str)])
-
+        network_interface_names: (Optional[list(str)]) A list of network interface names. The client will distribute the
+            connections across network interface names provided in this array. If any interface name is invalid, goes down,
+            or has any issues like network access, you will see connection failures.
+            This option is only supported on Linux, MacOS, and platforms that have either SO_BINDTODEVICE or IP_BOUND_IF. It
+            is not supported on Windows. `AWS_ERROR_PLATFORM_NOT_SUPPORTED` will be raised on unsupported platforms. On
+            Linux, SO_BINDTODEVICE is used and requires kernel version >= 5.7 or root privileges.
     """
 
     __slots__ = ('shutdown_event', '_region')
--- a/awscrt/s3.py
+++ b/awscrt/s3.py
@@ -237,8 +237,6 @@ class S3Client(NativeResource):
         assert isinstance(credential_provider, AwsCredentialsProvider) or credential_provider is None
         assert isinstance(tls_connection_options, TlsConnectionOptions) or tls_connection_options is None
         assert isinstance(part_size, int) or part_size is None
-        assert isinstance(network_interface_names, list) and all(isinstance(name, str)
-                                                                 for name in network_interface_names) or network_interface_names is None
 
         assert isinstance(
             throughput_target_gbps,
--- a/awscrt/s3.py
+++ b/awscrt/s3.py
@@ -207,7 +207,7 @@ class S3Client(NativeResource):
             between 2GiB and 8GiB (may change in future)
 
         network_interface_names: (Optional[list(str)]) A list of network interface names. The client will distribute the
-            connections across network interface names provided in this array. If any interface name is invalid, goes down,
+            connections across network interfaces provided in this list. If any interface name is invalid, goes down,
             or has any issues like network access, you will see connection failures.
             This option is only supported on Linux, MacOS, and platforms that have either SO_BINDTODEVICE or IP_BOUND_IF. It
             is not supported on Windows. `AWS_ERROR_PLATFORM_NOT_SUPPORTED` will be raised on unsupported platforms. On
--- a/awscrt/s3.py
+++ b/awscrt/s3.py
@@ -207,7 +207,7 @@ class S3Client(NativeResource):
             between 2GiB and 8GiB (may change in future)
 
         network_interface_names: (Optional[Sequence(str)]) A sequence of network interface names. The client will distribute the
-            connections across network interfaces provided. If any interface name is invalid, goes down,
+            connections across network interfaces. If any interface name is invalid, goes down,
             or has any issues like network access, you will see connection failures.
             This option is only supported on Linux, MacOS, and platforms that have either SO_BINDTODEVICE or IP_BOUND_IF. It
             is not supported on Windows. `AWS_ERROR_PLATFORM_NOT_SUPPORTED` will be raised on unsupported platforms. On
--- a/awscrt/s3.py
+++ b/awscrt/s3.py
@@ -293,8 +293,8 @@ class S3Client(NativeResource):
             throughput_target_gbps,
             enable_s3express,
             memory_limit,
-            s3_client_core,
-            network_interface_names)
+            network_interface_names,
+            s3_client_core)
 
     def make_request(
             self,
--- a/awscrt/s3.py
+++ b/awscrt/s3.py
@@ -205,6 +205,9 @@ class S3Client(NativeResource):
             client can use for buffering data for requests.
             Default values scale with target throughput and are currently
             between 2GiB and 8GiB (may change in future)
+
+        network_interface_names: (Optional[list(str)])
+
     """
 
     __slots__ = ('shutdown_event', '_region')
@@ -222,13 +225,17 @@ class S3Client(NativeResource):
             multipart_upload_threshold=None,
             throughput_target_gbps=None,
             enable_s3express=False,
-            memory_limit=None):
+            memory_limit=None,
+            network_interface_names=None):
         assert isinstance(bootstrap, ClientBootstrap) or bootstrap is None
         assert isinstance(region, str)
         assert isinstance(signing_config, AwsSigningConfig) or signing_config is None
         assert isinstance(credential_provider, AwsCredentialsProvider) or credential_provider is None
         assert isinstance(tls_connection_options, TlsConnectionOptions) or tls_connection_options is None
         assert isinstance(part_size, int) or part_size is None
+        assert isinstance(network_interface_names, list) and all(isinstance(name, str)
+                                                                 for name in network_interface_names) or network_interface_names is None
+
         assert isinstance(
             throughput_target_gbps,
             int) or isinstance(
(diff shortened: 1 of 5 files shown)


awscrt/s3.py
```diff
@@ -284,7 +295,8 @@ def on_shutdown():
             throughput_target_gbps,
             enable_s3express,
             memory_limit,
-            s3_client_core)
+            s3_client_core,
+            network_interface_names)
```
graebm, line 299: trivial/style: `s3_client_core` is kind of special, it's the only non-option-from-user thing passed along here.
Either keep it as the final arg, or move it to be the first arg so people aren't tempted to keep appending new args to the end of this list

source/s3_client.c
```diff
@@ -319,9 +322,22 @@ PyObject *aws_py_s3_client_new(PyObject *self, PyObject *args) {
...
 
+    struct aws_byte_cursor *network_interface_names = NULL;
+    int num_network_interface_names = 0;
+
+    if (network_interface_names_py != Py_None) {
+        if (!PyList_Check(network_interface_names_py)) {
+            PyErr_SetString(PyExc_TypeError, "Expected network_interface_names to be a list");
+            return NULL;
+        }
+        Py_ssize_t listSize = PyList_Size(network_interface_names_py);
```
graebm, line 333: Use the [Py_Sequence API](https://docs.python.org/3/c-api/sequence.html) so you can accept tuples too

awscrt/s3.py
```diff
@@ -222,13 +229,17 @@ def __init__(
...
+            memory_limit=None,
+            network_interface_names=None):
         assert isinstance(bootstrap, ClientBootstrap) or bootstrap is None
         assert isinstance(region, str)
         assert isinstance(signing_config, AwsSigningConfig) or signing_config is None
         assert isinstance(credential_provider, AwsCredentialsProvider) or credential_provider is None
         assert isinstance(tls_connection_options, TlsConnectionOptions) or tls_connection_options is None
         assert isinstance(part_size, int) or part_size is None
+        assert isinstance(network_interface_names, list) and all(isinstance(name, str)
+                                                                 for name in network_interface_names) or network_interface_names is None
```
graebm, line 241: you don't NEED to add these checks, since we check the type down in C, where we're going to accept lists OR tuples

we're not super consistent with these type-checks here in python code

source/s3_client.c
```diff
@@ -319,9 +322,22 @@ PyObject *aws_py_s3_client_new(PyObject *self, PyObject *args) {
...
+        num_network_interface_names = (size_t)listSize;
+    }
+
     struct s3_client_binding *s3_client = aws_mem_calloc(allocator, 1, sizeof(struct s3_client_binding));
 
     /* From hereon, we need to clean up if errors occur */
+    int result = AWS_OP_ERR;
```
graebm, line 340: remember how it makes me sad to have code that looks like it's doing aws-c-style error handling, when it's not
```suggestion
    bool success = false;
```

source/s3_client.c
```diff
@@ -319,9 +322,22 @@ PyObject *aws_py_s3_client_new(PyObject *self, PyObject *args) {
...
 
     PyObject *capsule = PyCapsule_New(s3_client, s_capsule_name_s3_client, s_s3_client_capsule_destructor);
     if (!capsule) {
```
graebm, line 343: after this if-check, there's still some manual cleanup and `return NULL` happening, change it to use the new `goto`

source/s3_client.c
```diff
@@ -319,9 +322,22 @@ PyObject *aws_py_s3_client_new(PyObject *self, PyObject *args) {
         signing_config = &default_signing_config;
     }
 
+    struct aws_byte_cursor *network_interface_names = NULL;
```
graebm, line 325: ugh. the code above this creates `anonymous_credentials` that needs cleanup if error occurs.

so like ... the location of the `* From hereon, we need to clean up if errors occur */` comment isn't accurate anymore

maybe just declare these higher up, and move the rest of the code for processing `network_interface_names_py` down in one place. Instead of splitting between 1 no-need-to-cleanup section and 1 yes-you-gotta-cleanup section

awscrt/s3.py
waahm7: Thanks, updated.

source/s3_client.c
waahm7: Thanks, updated to sequence.

source/s3_client.c
waahm7: Thanks, refactored the cleanup logic.

source/s3_client.c
waahm7: Thanks, I have tried to fix the cleanup logic and make the comment accurate.

source/s3_client.c
```diff
@@ -304,10 +307,16 @@ PyObject *aws_py_s3_client_new(PyObject *self, PyObject *args) {
 
     struct aws_signing_config_aws *signing_config = NULL;
     struct aws_credentials *anonymous_credentials = NULL;
+    struct aws_byte_cursor *network_interface_names = NULL;
+    int num_network_interface_names = 0;
```
graebm, line 311:
```suggestion
    size_t num_network_interface_names = 0;
```
right? you assign to this later like `num_network_interface_names = (size_t)listSize;`

source/s3_client.c
```diff
@@ -336,6 +342,29 @@ PyObject *aws_py_s3_client_new(PyObject *self, PyObject *args) {
     s3_client->py_core = py_core;
     Py_INCREF(s3_client->py_core);
 
+    if (network_interface_names_py != Py_None) {
+        if (!PySequence_Check(network_interface_names_py)) {
+            PyErr_SetString(PyExc_TypeError, "Expected network_interface_names to be a sequence.");
+            goto cleanup;
+        }
+        Py_ssize_t listSize = PySequence_Size(network_interface_names_py);
```
graebm, line 350: [docs say](https://docs.python.org/3/c-api/sequence.html#c.PySequence_Size)
> Returns ... -1 on failure

docs don't say whether a python exception was set in that case ... but [looking at the code](https://github.com/python/cpython/blob/376d66eb5080cf8076d767b0916c103463343963/Objects/abstract.c#L1609-L1610) seems like yeah there would be an exception

source/s3_client.c
```diff
@@ -336,6 +342,29 @@ PyObject *aws_py_s3_client_new(PyObject *self, PyObject *args) {
...
+        num_network_interface_names = (size_t)listSize;
+        network_interface_names =
+            aws_mem_calloc(allocator, num_network_interface_names, sizeof(struct aws_byte_cursor));
+        for (Py_ssize_t i = 0; i < num_network_interface_names; ++i) {
+            PyObject *strObj = PySequence_GetItem(network_interface_names_py, i);
```
graebm, line 355: Though python is generally excellent. You really need to read the docs for every single function in the c-api. Behavior is wildly inconsistent: [PyList_GetItem()](https://docs.python.org/3/c-api/list.html#c.PyList_GetItem) returned a borrowed reference, but `PySequence_GetItem()` returns a **new** reference.

I always put a `/* New reference */` or `/* Borrowed reference*/` comment after functions that return a `PyObject *`, to try and make this clear
```suggestion
            PyObject *strObj = PySequence_GetItem(network_interface_names_py, i); /* New reference */
```

or use the [PySequence_Fast](https://docs.python.org/3.11/c-api/sequence.html#c.PySequence_Fast) API, which will only do 1 new reference, instead of N (not that performance matters here, just do whatever makes the simplest cleanup code)

waahm7: Thanks, I have updated it with the comment and decRef. We can't use the fast API because https://docs.python.org/3.11/c-api/sequence.html#c.PySequence_Fast_GET_ITEM is not part of the stable API.

graebm: Why is PySequence_Fast() in the Stable ABI, but none of the functions you can call on that object are ?!? 🤷‍♀️

keep it weird Python

source/s3_client.c
```diff
@@ -336,6 +342,33 @@ PyObject *aws_py_s3_client_new(PyObject *self, PyObject *args) {
...
+        network_interface_names =
+            aws_mem_calloc(allocator, num_network_interface_names, sizeof(struct aws_byte_cursor));
+        for (size_t i = 0; i < num_network_interface_names; ++i) {
+            PyObject *str_obj = PySequence_GetItem(network_interface_names_py, i); /* New reference */
+            if (!str_obj) {
+                PyErr_SetString(PyExc_TypeError, "Expected network_interface_names elements to be non-null.");
+                goto cleanup;
+            }
+            network_interface_names[i] = aws_byte_cursor_from_pyunicode(str_obj);
+            Py_DECREF(str_obj);
```
graebm, line 364: 😬 now that I'm thinking about it, this isn't 1000% safe either

if (in theory) the sequence wasn't a tuple or list but something with a custom `__getitem__()` that returned a new object, then when we decref it here, it would get cleaned up, and our `char *` would be pointing at discarded memory

I looked at how PySequence_Fast_GET_ITEM() could get away with a borrowed reference, and it does it by ensuring a list or tuple was passed it, and if it was something else, it creates a list instead so it can be sure the items live at least as long as whatever's returned ([see code](https://github.com/python/cpython/blob/376d66eb5080cf8076d767b0916c103463343963/Objects/abstract.c#L2015))

anyway blugh ugh ugh ugh ugh I realize this bug exists elsewhere in our C-bindings

maybe keep these items alive until `cleanup`, similar to the array of cursors. Like:
- declare them before any gotos
    ```
    struct aws_byte_cursor *network_interface_name_cursors = NULL;
    PyObject **network_interface_name_pyobjects = NULL;
    ```
- `aws_mem_calloc()` them at the same time
- `network_interface_name_pyobjects[i] = PySequence_GetItem(network_interface_names_py, i); /* New reference */`
- then in cleanup, loop through and Py_XDECREF() them all before aws_mem_release of the tmp array

graebm: I'm not TRYING to make your life miserable. I'm just like "oh hey PySequence exists, let's use it" ... "uh oh"

Another option is doing like PySequence_Fast .. and turning it into a `list` (if necessary) before processing it. In python we could just be like:
```
if network_interface_names is not None:
    # ensure this is a list, so it's simpler to process in C
    if not isinstance(network_interface_names, list):
        network_interface_names = list(network_interface_names)
```

waahm7: Thanks, I have updated it to the second option.

awscrt/s3.py
```diff
@@ -235,6 +242,7 @@ def __init__(
             throughput_target_gbps,
             float) or throughput_target_gbps is None
         assert isinstance(enable_s3express, bool) or enable_s3express is None
+        assert isinstance(network_interface_names, Sequence) or network_interface_names is None
```
waahm7, line 245: I added the assert again because we will be converting it to a list.

graebm (approved):
what a long strange trip it's been

PR #579, opened 2024-07-09
fix free-bsd CI

TingDaoK:
*Issue #, if available:*

- FreeBSD 14:latest dropped python3.9, according to [here](https://www.freshports.org/devel/py-pip/) 

*Description of changes:*

- install py-pip via `pkg install devel/py-pip`. https://www.freshports.org/devel/py-pip/
- install py-urllib3 via `pkg install net/py-urllib3`. https://www.freshports.org/net/py-urllib3/
- install py-awscli via `pkg install devel/py-awscli`. https://www.freshports.org/devel/py-awscli/

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -232,7 +232,7 @@ jobs:
           shell: bash
           environment_variables: AWS_ACCESS_KEY_ID AWS_SECRET_ACCESS_KEY AWS_DEFAULT_REGION AWS_REGION
           run: |
-            sudo pkg install -y python39 py39-pip py39-urllib3 py39-pip py39-awscli cmake
+            sudo pkg install -y python3 py-pip py-urllib3 py-awscli cmake
             python3 -c "from urllib.request import urlretrieve; urlretrieve('${{ env.BUILDER_HOST }}/${{ env.BUILDER_SOURCE }}/${{ env.BUILDER_VERSION }}/builder.pyz', 'builder')"
             chmod a+x builder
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -232,7 +232,7 @@ jobs:
           shell: bash
           environment_variables: AWS_ACCESS_KEY_ID AWS_SECRET_ACCESS_KEY AWS_DEFAULT_REGION AWS_REGION
           run: |
-            sudo pkg install -y python3 py39-urllib3 py39-pip py39-awscli cmake
+            sudo pkg install -y python3 devel/py-pip net/py-urllib3 devel/py-awscli cmake
```
graebm, line 235: did `install python3` change from being 3.9 to 3.11?

does `python3` point to whatever is "latest"? and like `devel/py-pip` points to whatever is "latest"?

If so, this seems fine

graebm (approved):

.github/workflows/ci.yml
TingDaoK: not sure if they just updated recently or not. but the current default version for python is 3.11 https://cgit.freebsd.org/ports/tree/Mk/bsd.default-versions.mk#n139

https://wiki.freebsd.org/Python looks like they will use the default version for all the python packages.

TingDaoK: https://github.com/freebsd/freebsd-ports/commit/62a23ef0936e21fef507653054d6732bb7ef085f oh, it was updated 2 month ago.

PR #587, opened 2024-08-22
Add Python 3.13 CI

waahm7:
- Adds CI for Python 3.13
- Updated CD Pipeline to use Python 3.13 stable API due to deprecated function
- Moves our CI from hard coded creds to temporary creds
- Fixes some MQTT tests that were relying on hard coded creds in the environment.

### DIFF
--- a/source/http_stream.c
+++ b/source/http_stream.c
@@ -203,7 +203,7 @@ static void s_on_stream_complete(struct aws_http_stream *native_stream, int erro
     }
 
     /* DECREF python self, we don't need to force it to stay alive any longer. */
-    Py_DECREF(PyWeakref_GetObject(stream->self_proxy));
+    Py_XDECREF(stream->self_proxy);
 
     PyGILState_Release(state);
     /*************** GIL RELEASE ***************/
--- a/source/http_stream.c
+++ b/source/http_stream.c
@@ -205,11 +205,12 @@ static void s_on_stream_complete(struct aws_http_stream *native_stream, int erro
     /* DECREF python self, we don't need to force it to stay alive any longer. */
     PyObject *self = Py_None;
 #if PY_VERSION_HEX >= 0x030D0000 /* Check if Python version is 3.13 or higher */
+    int result_getref = PyWeakref_GetRef(stream->self_proxy, &self);/* strong reference */
     /* Ignore error, stream is already complete */
-    PyWeakref_GetRef(stream->self_proxy, &self) < 0);/* strong reference */
+    (void) result_getref;
 #else
     /* PyWeakref_GetObject is deprecated since python 3.13 */
-    PyWeakref_GetObject(stream->self_proxy); /* borrowed reference */
+    self = PyWeakref_GetObject(stream->self_proxy); /* borrowed reference */
 #endif
     Py_XDECREF(self);
 #if PY_VERSION_HEX >= 0x030D0000 /* Check if Python version is 3.13 or higher */
--- a/source/http_stream.c
+++ b/source/http_stream.c
@@ -204,10 +204,10 @@ static void s_on_stream_complete(struct aws_http_stream *native_stream, int erro
 
     /* DECREF python self, we don't need to force it to stay alive any longer. */
     PyObject *self = Py_None;
-#if PY_VERSION_HEX >= 0x030D0000 /* Check if Python version is 3.13 or higher */
-    int result_getref = PyWeakref_GetRef(stream->self_proxy, &self);/* strong reference */
+#if PY_VERSION_HEX >= 0x030D0000                                     /* Check if Python version is 3.13 or higher */
+    int result_getref = PyWeakref_GetRef(stream->self_proxy, &self); /* strong reference */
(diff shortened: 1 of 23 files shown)


source/http_stream.c
```diff
@@ -203,7 +203,19 @@ static void s_on_stream_complete(struct aws_http_stream *native_stream, int erro
...
+    int result_getref = PyWeakref_GetRef(stream->self_proxy, &self); /* strong reference */
+    /* Ignore error, stream is already complete */
+    (void)result_getref;
+#else
+    /* PyWeakref_GetObject is deprecated since python 3.13 */
+    self = PyWeakref_GetObject(stream->self_proxy); /* borrowed reference */
+#endif
+    Py_XDECREF(self);
+#if PY_VERSION_HEX >= 0x030D0000 /* Check if Python version is 3.13 or higher */
+    Py_XDECREF(self);
```
DmitriyMusatkin, line 217: you end up decreffing it twice in 3.13 branch?

waahm7: Yes, one is the original decref that we need to do, and the second decref is for the strong reference that the new API returned.

source/mqtt_client_connection.c
```diff
@@ -227,7 +271,17 @@ static void s_on_connection_resumed(
...
+    PyObject *self = Py_None;
+#if PY_VERSION_HEX >= 0x030D0000                                  /* Check if Python version is 3.13 or higher */
+    if (PyWeakref_GetRef(py_connection->self_proxy, &self) < 0) { /* strong reference */
+        PyErr_WriteUnraisable(PyErr_Occurred());
+        goto on_done;
+    }
+#else
+    /* PyWeakref_GetObject is deprecated since python 3.13 */
+    self = PyWeakref_GetObject(py_connection->self_proxy); /* borrowed reference */
+#endif
```
graebm, line 283: the logic is copy/pasted a lot of time. How about backporting PyWeakref_GetRef to older versions

module.h:
```
#if PY_VERSION_HEX < 0x30D0000
// PyWeakref_GetRef() was added in Python3.13, make it available in older versions
int PyWeakref_GetRef(PyObject *ref, PyObject **pobj);
#endif
```

module.c:
```
#if PY_VERSION_HEX < 0x30D0000
// PyWeakref_GetRef() was added in Python3.13, make it available in older versions
int PyWeakref_GetRef(PyObject *ref, PyObject **pobj) {
    ...
}
#endif
```

graebm: https://github.com/python/cpython/blob/a9c2bc16349c2be3005f97249f3ae9699988f218/Objects/weakrefobject.c#L931-L947

waahm7: Not sure if we can backport it. `PyWeakref_GetRef` depends upon the internal function `_PyWeakref_GetRef`, which is probably located at https://fossies.org/linux/misc/Python-3.13.0rc2.tar.xz/Python-3.13.0rc2/Include/internal/pycore_weakref.h, and that function also depends upon several internal functions.

We can potentially add `PyWeakref_GetRef`, which calls `PyWeakref_GetObject` and also increments the ref count, so that we can avoid the copy-paste.

waahm7: Thanks, updated. I have created a new function which abstracts the differences between APIs into a single place, and we don't need to do any code duplication. I didn't define `PyWeakref_GetRef` for pre-Python 3.13 because I thought it would be confusing for future readers that we are calling the Python 3.13+ function, and having our own definition for pre-Python 3.13 which does something slightly different.

source/module.c
```diff
@@ -516,6 +516,27 @@ PyObject *aws_py_memory_view_from_byte_buffer(struct aws_byte_buf *buf) {
...
 
+PyObject *aws_py_weakref_get_ref(PyObject *object) {
+    PyObject *self = Py_None;
+#if PY_VERSION_HEX >= 0x030D0000               /* Check if Python version is 3.13 or higher */
+    if (PyWeakref_GetRef(object, &self) < 0) { /* strong reference */
+        return Py_None;
+    }
+#else
+    /* PyWeakref_GetObject is deprecated since python 3.13 */
+    self = PyWeakref_GetObject(object); /* borrowed reference */
```
graebm, line 527: I vote we just have it behave like the new-function in ALL versions and return a New reference

Otherwise we live in a confusing world where it behaves 2 different ways, based on the version of python that built the wheel (not necessarily the version of python the user is running)

source/module.c
```diff
@@ -516,6 +516,27 @@ PyObject *aws_py_memory_view_from_byte_buffer(struct aws_byte_buf *buf) {
...
+    return self;
+}
+
+void aws_py_weakref_release_ref(PyObject *object) {
+    (void)object;
+    /* Python versions before 3.13 returns a borrowed reference */
+#if PY_VERSION_HEX >= 0x030D0000
+    Py_XDECREF(object);
+#endif
+}
```
graebm, line 538: and then we don't need this function at all

graebm (approved):

PR #594, opened 2024-09-10
Update setup.py

gorloffslava:
This PR adds a new buildtime environment variable `AWS_CRT_BUILD_USE_SHARED_LIBS`. 

When it's set to `1`, it overrides the default behavior of trying to link static library variants on Unix OSs (except `darwin` ). 

This default behavior is enforced by the following lines:
https://github.com/awslabs/aws-crt-python/blob/2dae492d57b40c68839f2ecd7867bea34e6f9f1a/setup.py#L340-L348

Currently, there is no way to disable it and prefer shared libraries instead, what this PR solves.

As a continuation of the [#593](https://github.com/awslabs/aws-crt-python/pull/593), it allows to reduce the libraries footprint in disk-restricted environments like Lambda, promote shared libraries reuse in RAM (e.g., if multiple Python packages depend on them), and generally make it more aligned with best native Python extensions build practices.

This PR is backward-compatible as feature it provides is opt-in, it doesn't change the default behavior of preferring static libraries.
 
### DIFF
--- a/setup.py
+++ b/setup.py
@@ -138,6 +138,9 @@ def using_system_libs():
     """If true, don't build any dependencies. Use the libs that are already on the system."""
     return os.getenv('AWS_CRT_BUILD_USE_SYSTEM_LIBS') == '1'
 
+def using_shared_libs():
+    """If true, the shared libs are used instead of the static ones, which is the default. Has no effect on Windows and Darwin."""
+    return os.getenv('AWS_CRT_BUILD_USE_SHARED_LIBS') == '1'
 
 def using_system_libcrypto():
     """If true, don't build AWS-LC. Use the libcrypto that's already on the system."""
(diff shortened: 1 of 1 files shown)


graebm:
Thanks for the PR. I'm looking into this ... just busy

BwL1289:
No problem. Apologies I missed this in the previous issue/PR.

graebm:
Here's my take on it: https://github.com/awslabs/aws-crt-python/pull/596

It's the opposite of your approach. I'll fix AWS's internal build system to use the new-extra-weird env-var. Instead of forcing slightly-less-weird users like yourself to set 2 env-vars.

graebm:
Thoughts on #596? Does this work for you?

BwL1289:
@graebm just seeing - will test and get back to you. Thank you!

PR #596, opened 2024-09-12
AWS_CRT_BUILD_FORCE_STATIC_LIBS

graebm:
**Issue:**
https://github.com/awslabs/aws-crt-python/pull/593 introduced env-var, `AWS_CRT_BUILD_USE_SYSTEM_LIBS=1`, but it didn't work unless those system libs were built statically.

**Diagnosis:**
The reason it didn't work is: there's a hack in [setup.py](https://github.com/awslabs/aws-crt-python/blob/2dae492d57b40c68839f2ecd7867bea34e6f9f1a/setup.py#L339-L348) that forces dependencies to be linked statically on Unix variants (excluding macOS). It worked when I tested on macOS because the hack isn't applied there.

The reason for the hack is: in Brazil (internal AWS build system), dependencies are available as both static and dynamic libs. But we prefer to link the static ones so that, if a python application is packaged for AWS lambda, the developer has fewer files they need to chase down and copy into their .zip package.

**Description of changes:**
Don't force static libs to be used unless `AWS_CRT_BUILD_FORCE_STATIC_LIBS=1` (new env-var).  Otherwise (by default), the linker will use whatever version of the lib is available (if both are available, linkers generally prefer dynamic). This only applies to non-OS dependencies, we won't force libc to be linked statically.

It's unlikely this env-var will be used anywhere except Brazil.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -136,12 +136,18 @@ def get_cmake_path():
 
 def using_system_libs():
     """If true, don't build any dependencies. Use the libs that are already on the system."""
-    return os.getenv('AWS_CRT_BUILD_USE_SYSTEM_LIBS') == '1'
+    return (os.getenv('AWS_CRT_BUILD_USE_SYSTEM_LIBS') == '1'
+            or not os.path.exists(os.path.join(PROJECT_DIR, 'crt', 'aws-c-common', 'CMakeLists.txt')))
 
 
 def using_system_libcrypto():
     """If true, don't build AWS-LC. Use the libcrypto that's already on the system."""
     return using_system_libs() or os.getenv('AWS_CRT_BUILD_USE_SYSTEM_LIBCRYPTO') == '1'
+
+
+def forcing_static_libs():
+    """If true, force libs to be linked statically."""
+    return os.getenv('AWS_CRT_BUILD_FORCE_STATIC_LIBS') == '1'
 
 
 class AwsLib:
@@ -152,12 +158,11 @@ class AwsLib:
 
 
 # The extension depends on these libs.
-# They're built along with the extension.
+# They're built along with the extension (unless using_system_libs() is True)
 AWS_LIBS = []
 if sys.platform != 'darwin' and sys.platform != 'win32':
-    if not using_system_libcrypto():
-        # aws-lc produces libcrypto.a
-        AWS_LIBS.append(AwsLib('aws-lc', libname='crypto'))
+    # aws-lc produces libcrypto.a
+    AWS_LIBS.append(AwsLib('aws-lc', libname='crypto'))
     AWS_LIBS.append(AwsLib('s2n'))
 AWS_LIBS.append(AwsLib('aws-c-common'))
 AWS_LIBS.append(AwsLib('aws-c-sdkutils'))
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -337,38 +340,40 @@ def awscrt_ext():
...
+        # (If AWS-LC's CMakeLists.txt removes these lines in the future, we can remove this hack here as well)
+        if sys.platform.startswith('openbsd'):
+            if not using_system_libcrypto():
                 extra_link_args += ['-Wl,--no-execute-only']
 
         # FreeBSD doesn't have execinfo as a part of libc like other Unix variant.
         # Passing linker flag to link execinfo properly
         if sys.platform.startswith('freebsd'):
-            extra_link_args += ['-lexecinfo']
+            libraries += ['execinfo']
```
graebm, line 376: this wasn't broken, it was just weird we were adding this library via `extra_link_args` instead of via `libraries`

setup.py
```diff
@@ -337,38 +340,40 @@ def awscrt_ext():
...
-            extra_link_args += ['-Wl,--exclude-libs,libcrypto.a']
-
-            # OpenBSD 7.4+ defaults to linking with --execute-only, which is bad for AWS-LC.
-            # See: https://github.com/aws/aws-lc/blob/4b07805bddc55f68e5ce8c42f215da51c7a4e099/CMakeLists.txt#L44-L53
-            # (If AWS-LC's CMakeLists.txt removes these lines in the future, we can remove this hack here as well)
-            if sys.platform.startswith('openbsd'):
+        # hide the symbols from libcrypto.a
+        # this prevents weird crashes if an application also ends up using
+        # libcrypto.so from the system's OpenSSL installation.
+        # Do this even if using system libcrypto, since it could still be a static lib.
```
graebm, line 363: I moved this out of the if-statement because:
> Do this even if using system libcrypto, since it could still be a static lib.

DmitriyMusatkin (approved):

setup.py
```diff
@@ -152,12 +158,11 @@ def __init__(self, name, extra_cmake_args=[], libname=None):
...
 # The extension depends on these libs.
-# They're built along with the extension.
+# They're built along with the extension (unless using_system_libs() is True)
 AWS_LIBS = []
 if sys.platform != 'darwin' and sys.platform != 'win32':
-    if not using_system_libcrypto():
-        # aws-lc produces libcrypto.a
-        AWS_LIBS.append(AwsLib('aws-lc', libname='crypto'))
+    # aws-lc produces libcrypto.a
+    AWS_LIBS.append(AwsLib('aws-lc', libname='crypto'))
```
graebm, line 165: I did this so that aws-lc is in `AWS_LIBS` even if we're not building it, in the same way that **all the other non-OS libs** are in `AWS_LIBS` even if we're not building them. This removes some weirdness later where we had to add `crypto` back into the list of libraries we need to link against

graebm:
Oof bad timing. Some other recent build changes ([1](https://github.com/awslabs/aws-crt-python/pull/603), [2](https://github.com/awslabs/aws-crt-python/pull/584), [3](https://github.com/awslabs/aws-crt-python/pull/602)) are making things rocky for our team this last week. Once those issues settle down, I'll work on merging this

graebm:
OK, it's out in [v0.23.3](https://github.com/awslabs/aws-crt-python/releases/tag/v0.23.3)

PR #598, opened 2024-09-26
Fix Release Pipeline for distutils

waahm7:
*Description of changes:*
- Disutils is deprecated, however, currently there is no alternative see: https://github.com/pypa/setuptools/issues/2806
- This PR restricts disutils dependency to Windows only
- Drop Python 3.7 as it is EOL 
- Install requirements-dev.txt in Python 3.13 since manylinux images removed it see: https://github.com/pypa/manylinux/pull/1564/files#diff-de469cc9cf8992ffaec661b4e087cfded3ec6da722072ec1fdf11f61690fd1b8R3

### DIFF
--- a/continuous-delivery/build-wheels-manylinux2014-x86_64.sh
+++ b/continuous-delivery/build-wheels-manylinux2014-x86_64.sh
@@ -4,9 +4,6 @@ set -ex
 
 /opt/python/cp39-cp39/bin/python ./continuous-delivery/update-version.py
 
-/opt/python/cp37-cp37m/bin/python setup.py sdist bdist_wheel
-auditwheel repair --plat manylinux2014_x86_64 dist/awscrt-*cp37*.whl
-
 /opt/python/cp38-cp38/bin/python setup.py sdist bdist_wheel
 auditwheel repair --plat manylinux2014_x86_64 dist/awscrt-*cp38*.whl
 
--- a/continuous-delivery/build-wheels-manylinux2014-x86_64.sh
+++ b/continuous-delivery/build-wheels-manylinux2014-x86_64.sh
@@ -20,6 +20,10 @@ auditwheel repair --plat manylinux2014_x86_64 dist/awscrt-*cp311*.whl
 # The 3.11 wheel uses the stable ABI, so it works with newer versions too.
 
 # We are using the Python 3.13 stable ABI from Python 3.13 onwards because of deprecated functions.
+# Manylinux images don't contain setuptools from Python 3.13, so we need to install it.
+# Install in a custom location due to access issues.
+/opt/python/cp313-cp313/bin/python -m pip install --target ./local -r requirements-dev.txt
+export PYTHONPATH=./local:$PYTHONPATH
 /opt/python/cp313-cp313/bin/python setup.py sdist bdist_wheel
 auditwheel repair --plat manylinux2014_x86_64 dist/awscrt-*cp313*.whl
 
--- a/setup.py
+++ b/setup.py
@@ -75,6 +75,7 @@ def determine_generator_args():
     if sys.platform == 'win32':
         try:
             # See which compiler python picks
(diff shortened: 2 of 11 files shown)


setup.py
```diff
@@ -326,6 +327,11 @@ def awscrt_ext():
     libraries.reverse()
 
     if sys.platform == 'win32':
+        # distutils is deprecated in Python 3.10 and removed in 3.12. However, it still works because Python defines a compatibility interface as long as setuptools is installed.
+        # We don't have an official alternative for distutils.ccompiler as of September 2024. See: https://github.com/pypa/setuptools/issues/2806
+        # Once that issue is resolved, we can migrate to the official solution.
+        # For now, restrict distutils to Windows only, where it's needed.
+        import distutils.ccompiler
```
TingDaoK, line 334: Can we move the import to the beginning of the file instead? So that we don't need to have two place to import the package?

setup.py
```diff
@@ -305,8 +304,10 @@ def run(self):
 class bdist_wheel_abi3(bdist_wheel):
     def get_tag(self):
         python, abi, plat = super().get_tag()
-        if python.startswith("cp") and sys.version_info >= (3, 11):
-            # on CPython, our wheels are abi3 and compatible back to 3.11
+        # on CPython, our wheels are abi3 and compatible back to 3.11
+        if python.startswith("cp") and sys.version_info >= (3, 13):
+            return "cp313", "abi3", plat
```
TingDaoK, line 309: worth to mention that cp313 has a different abi3 (Not sure if it's right wording.)?

continuous-delivery/build-wheels-manylinux2014-x86_64.sh
```diff
@@ -23,6 +20,10 @@ auditwheel repair --plat manylinux2014_x86_64 dist/awscrt-*cp311*.whl
 # The 3.11 wheel uses the stable ABI, so it works with newer versions too.
 
 # We are using the Python 3.13 stable ABI from Python 3.13 onwards because of deprecated functions.
+# Manylinux images don't contain setuptools from Python 3.13, so we need to install it.
+# Install in a custom location due to access issues.
+/opt/python/cp313-cp313/bin/python -m pip install --target ./local -r requirements-dev.txt
+export PYTHONPATH=./local:$PYTHONPATH
```
TingDaoK, line 26: (debatable) Move those to a `prepare-manylinux.sh`.
So that have one source instead of copy/paste to 4 places?
And we will remove manylinux1 sometime soon, so it's fine that not invoking it from manylinux1?
More future prove?

TingDaoK (approved):
fix&ship

setup.py
waahm7: If we move the import to the beginning of the file, it will be imported everywhere. Because this is only required on Windows and is deprecated, I have restricted it to Windows.

setup.py
waahm7: Thanks, we already mention it at https://github.com/awslabs/aws-crt-python/blob/main/setup.py#L408. I will add a comment here as well.

continuous-delivery/build-wheels-manylinux2014-x86_64.sh
waahm7: I don't see any prepare-manylinux.sh script. Do you mean creating a new one? I think it's fine for now, since it's just two lines of code, and the rest of the code is different per platform. This is also ABI compatible wheel so we don't expect to do this often.

manylinux1 doesn't support Python 3.13, I believe, so we don't need to add it there.

setup.py
TingDaoK: I mean move the
```
if sys.platform == 'win32':
   import distutils.ccompiler
```
to the top of the file

waahm7: Sorry, missed it in this PR. I have created #599 to fix this.

Issue #604, opened 2024-10-18
aws-c-s3 upgrade 0.6.7 -> 0.6.9 s3 test fail


thomas-roos:
### Describe the bug

ERROR: test_sanity_network_interface_names (test.test_s3.S3ClientTest.test_sanity_network_interface_names)                                                                                                                                                                                
----------------------------------------------------------------------                                                                                                                                                                                                                    
Traceback (most recent call last):                                                                                                                                                                                                                                                        
  File "/usr/lib/aws-crt-python/ptest/tests/test/test_s3.py", line 225, in test_sanity_network_interface_names                                                                                                                                                                            
    s3_client = s3_client_new(True, self.region, network_interface_names=("eth0", "eth1"))                                                                                                                                                                                                
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^                                                                                                                                                                                                
  File "/usr/lib/aws-crt-python/ptest/tests/test/test_s3.py", line 183, in s3_client_new                                                                                                                                                                                                  
    s3_client = S3Client(                                                                                                                                                                                                                                                                 
                ^^^^^^^^^                                                                                                                                                                                                                                                                 
  File "/usr/lib/aws-crt-python/ptest/tests/awscrt/s3.py", line 288, in __init__                                                                                                                                                                                                          
    self._binding = _awscrt.s3_client_new(                                                                                                                                                                                                                                                
                    ^^^^^^^^^^^^^^^^^^^^^^                                                                                                                                                                                                                                                
RuntimeError: 34 (AWS_ERROR_INVALID_ARGUMENT): An invalid argument was passed to a function.    

### Expected Behavior

worked as with 0.6.7

### Current Behavior

ERROR: test_sanity_network_interface_names (test.test_s3.S3ClientTest.test_sanity_network_interface_names)                                                                                                                                                                                
----------------------------------------------------------------------                                                                                                                                                                                                                    
Traceback (most recent call last):                                                                                                                                                                                                                                                        
  File "/usr/lib/aws-crt-python/ptest/tests/test/test_s3.py", line 225, in test_sanity_network_interface_names                                                                                                                                                                            
    s3_client = s3_client_new(True, self.region, network_interface_names=("eth0", "eth1"))                                                                                                                                                                                                
[...]

PR #605, opened 2024-10-18
Fix Network Interface Test

waahm7:
*Issues: *
https://github.com/awslabs/aws-crt-python/issues/604

*Description of changes:*
- Aws-c-s3 now validates network interface names at client initialization. 

### DIFF
--- a/test/test_s3.py
+++ b/test/test_s3.py
@@ -222,6 +222,7 @@ class S3ClientTest(NativeResourceTest):
         self.assertIsNotNone(s3_client)
 
     def test_sanity_network_interface_names(self):
+        # This is just a sanity test to ensure that we are passing the parameter correctly.
         with self.assertRaises(Exception):
             s3_client_new(True, self.region, network_interface_names=("eth0", "invalid-network-interface"))
 
--- a/test/test_s3.py
+++ b/test/test_s3.py
@@ -221,6 +221,10 @@ class S3ClientTest(NativeResourceTest):
         s3_client = s3_client_new(True, self.region)
         self.assertIsNotNone(s3_client)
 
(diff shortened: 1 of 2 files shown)


test/test_s3.py
```diff
@@ -222,8 +222,8 @@ def test_sanity_secure(self):
         self.assertIsNotNone(s3_client)
 
     def test_sanity_network_interface_names(self):
-        s3_client = s3_client_new(True, self.region, network_interface_names=("eth0", "eth1"))
-        self.assertIsNotNone(s3_client)
+        with self.assertRaises(Exception):
+            s3_client_new(True, self.region, network_interface_names=("eth0", "invalid-network-interface"))
```
TingDaoK, line 226: should we also check the happy case?

waahm7: It's not easy to test since only a few platforms have a happy case, and even then, it's different for each platform. This test is only verifying that we pass the value to c-s3 correctly.

TingDaoK: We can add a comment here that the functionality was tested in C properly, and here we make sure the binding works.

TingDaoK (approved):

test/test_s3.py
waahm7: I think the term "sanity" in the test name already signifies this intent. I can add a comment if you still think we should include one here.

TingDaoK: Well, we have bunch of sanity in the name of the test in aws-c-*, they are meant to simply test it works in the happy case,. eg: https://github.com/awslabs/aws-c-http/blob/main/tests/CMakeLists.txt#L410

I don't think we have a fairly common agreement on "sanity" in the test name means.

It's fine to me. But just mentioning.

PR #606, opened 2024-10-21
Update Win Sanity Check Script 

waahm7:
*Issue #, if available:*

*Description of changes:*
- Update win sanity check script to not use python 37


DmitriyMusatkin (approved):

TingDaoK (commented):
https://github.com/awslabs/aws-crt-python/blob/main/codebuild/cd/manylinux-x86-build.yml#L11
https://github.com/awslabs/aws-crt-python/blob/main/codebuild/cd/manylinux-x64-build.yml#L11

should we update those as well?

PR #613, opened 2024-12-11
Fix macos universal2 detection in setup.py

graebm:
Thanks @TingDaoK for discovering this copy/paste bug.

This wasn't causing any actual problems (other than longer-than-necessary build times on macos-but-not-universal2 builds of python ... which are exceedingly rare these days), but it's fixed now.


TingDaoK (approved):
🚢

PR #626, opened 2025-02-28
Fix aarch64 runners

DmitriyMusatkin:
*Issue #, if available:*

*Description of changes:*
We rely on qemu to test aarch64 on x86-64 github runners. 
Qemu had a bug where they had a bug with incorrectly mapping address space. That mostly worked fine until kernel patch that increased entropy in their address randomization. TLDR; qemu likes to segfault a lot against new kernels on our aarch64 test. Solution is to switch to a different github qemu user static image that forces a latest version of qemu, that has the bug fixed.
if you want to read more - https://github.com/tonistiigi/binfmt/issues/240

in a separate issue setuptools now prefers systems version of packages, which breaks against packaging thats on the ubuntu runners by defaults, so add packaging to our deps to ensure its getting the latest version. 

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -125,7 +125,7 @@ jobs:
 
       # set arm arch
       - name: Install qemu/docker
-        run: docker run --privileged --rm tonistiigi/binfmt --install arm/v7
+        run: docker run --privileged --rm tonistiigi/binfmt --install armv7
 
       - name: Build ${{ env.PACKAGE_NAME }}
         run: |
@@ -201,7 +201,7 @@ jobs:
         run: |
           python3 -c "from urllib.request import urlretrieve; urlretrieve('${{ env.BUILDER_HOST }}/${{ env.BUILDER_SOURCE }}/${{ env.BUILDER_VERSION }}/builder.pyz?run=${{ env.RUN }}', 'builder')"
           chmod a+x builder
-          ./builder build -p ${{ env.PACKAGE_NAME }} --python=python3.11
+          ./builder build -p ${{ env.PACKAGE_NAME }} --python=python3.12
       - name: Assert libcrypto.so used
         run: |
           # assert it's linked against the system's libcrypto.so
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -125,7 +125,7 @@ jobs:
 
       # set arm arch
       - name: Install qemu/docker
-        run: docker run --privileged --rm tonistiigi/binfmt --install all
+        run: docker run --privileged --rm tonistiigi/binfmt --install arm/v7
 
       - name: Build ${{ env.PACKAGE_NAME }}
         run: |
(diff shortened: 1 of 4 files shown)


.github/workflows/ci.yml
```diff
@@ -67,15 +67,15 @@ jobs:
           aws-region: ${{ env.AWS_DEFAULT_REGION }}
       # Only aarch64 needs this, but it doesn't hurt anything
       - name: Install qemu/docker
-        run: docker run --rm --privileged multiarch/qemu-user-static --reset -p yes
+        run: docker run --privileged --rm tonistiigi/binfmt --install all
```
waahm7, line 70: Should we just do --install arm64 instead of all?

waahm7 (approved):

.github/workflows/ci.yml
waahm7: Nevermind, I see that you tried it.

DmitriyMusatkin: yeah, good point. updated

.github/workflows/ci.yml
```diff
@@ -125,18 +125,19 @@ jobs:
 
       # set arm arch
       - name: Install qemu/docker
-        run: docker run --rm --privileged multiarch/qemu-user-static --reset -p yes
+        run: docker run --privileged --rm tonistiigi/binfmt --install all
```
waahm7, line 128: Any reason for all? The image architecture is arm-v7 https://github.com/awslabs/aws-crt-builder/blob/42d15ba4ee3fbe0e812e25094a8cf59eaef0cf1e/.github/docker-images/raspbian-bullseye/Dockerfile#L1.

.github/workflows/ci.yml
```diff
@@ -185,7 +186,7 @@ jobs:
           ./linux-container-ci.sh ${{ env.BUILDER_VERSION }} aws-crt-${{ env.LINUX_BASE_IMAGE }} build -p ${{ env.PACKAGE_NAME }} --compiler=${{ matrix.compiler }}
 
   use-system-libcrypto:
-    runs-on: ubuntu-20.04 # latest
+    runs-on: ubuntu-22.04 # latest
```
waahm7, line 189: Why 22?

DmitriyMusatkin: 24 is openssl3.0. 22 is still 1.1.1

.github/workflows/ci.yml
DmitriyMusatkin: ah, i was being lazy :) 
let me set it to armv7

.github/workflows/ci.yml
waahm7: Maybe update the comment from # latest to explain the OpenSSL difference.

waahm7 (approved):

.github/workflows/ci.yml
DmitriyMusatkin: actually, nvm. looks like 22 is now bundling openssl3.0 as well. lets switch to 24 then, c-cal still provides coverage for 1.1.1.

PR #628, opened 2025-03-11
Fork workaround

TingDaoK:
*Issue #, if available:*

- We don't support fork and we probably will not support fork any time soon
- As the current user, if you have to use fork, provide some caveat and a test as an example how you can make it work.
- Updated submodules to bring https://github.com/awslabs/aws-c-common/pull/1192
```
aws-c-auth         v0.8.6 -> v0.8.7
aws-c-common       v0.12.0 -> v0.12.1
aws-c-http         v0.9.3 -> v0.9.5
aws-c-s3           v0.7.12 -> v0.7.13
aws-lc             v1.48.1 -> v1.48.5
```


README.md
```diff
@@ -83,4 +84,22 @@ static: certificate has an existing certificate-key pair that was previously imp
...
+But, as the CRT user and you **really** handles threads carefully enough, you could make it work.
+
+The major issue for fork is the forked process will only have the thread invokes the fork. All other threads will be gone from the child process.
+To workaround this:
+1. Release all CRT resources with background threads. Everything depends on `io.EventLoopGroup`, and the `io.EventLoopGroup` object itself, will needs to be cleaned up before fork
+2. Wait for all threads to join before fork. `common.join_all_native_threads()` is a helper to track all the threads created by CRT will be joined from the main thread.
+
+Specifically, refer to `test.test_s3.py.S3RequestTest.test_fork_workaround` for an example of the workaround.
+
+More caveat: MacOS system libraries may start threads. You may see random crashes even if all CRT threads joined before fork on Mac. Try not using fork on MacOS, it's not a default for python since 3.8, more context can be found [here](https://docs.python.org/3/library/multiprocessing.html#contexts-and-start-methods).
```
graebm, line 105: here's a rewrite I agonized over
```
## Fork and Multiprocessing

aws-crt-python uses background threads. This makes [os.fork()](https://docs.python.org/3/library/os.html#os.fork) unsafe. In a forked child process, all background threads vanish. The child will hang or crash when it tries to communicate with any of these (vanished) threads.

Unfortunately, Python's [multiprocessing](https://docs.python.org/3/library/multiprocessing.html) module defaults to using fork when it creates child processes (on POSIX systems except macOS, in Python versions 3.13 and earlier). `multiprocessing` is used under the hood by many tools that do work in parallel, including [concurrent.futures.ProcessPoolExecutor](https://docs.python.org/3/library/concurrent.futures.html#concurrent.futures.ProcessPoolExecutor), and [pytorch.multiprocessing](https://pytorch.org/docs/stable/multiprocessing.html).

If you need to use `multiprocessing` with aws-crt-python, set it to use "spawn" or "forkserver" instead of "fork" ([see docs](https://docs.python.org/3/library/multiprocessing.html#contexts-and-start-methods)). The Python community agrees, and `multiprocessing` will changes its default from "fork" to "spawn" in 3.14. It already uses "spawn" by default on macOS (because system libraries may start threads) and on Windows (because fork does not exist).

If you **must** use fork with aws-crt-python, you may be able to avoid hangs and crashes if you manage your threads very carefully:

1.	Release all CRT resources with background threads (e.g. clean up any `io.EventLoopGroup` instances).
2.	Join all CRT threads before forking (use `common.join_all_native_threads()`).

For an example, see `test.test_s3.py.S3RequestTest.test_fork_workaround`.
```

README.md
```diff
@@ -45,15 +45,16 @@ To simplify installation, aws-crt-python has its own copy of libcrypto.
 This lets you install a wheel from PyPI without having OpenSSL installed.
 Unix wheels on PyPI come with libcrypto statically compiled in.
 Code to build libcrypto comes from [AWS-LC](https://github.com/aws/aws-lc).
-AWS-LC's code is included in the PyPI source package,
+AWS-LC's code is included in the PyPI source package, 
 and the git repository includes it as a submodule.
 
-If you need aws-crt-python to use the libcrypto included on your system,
+If you need aws-crt-python to use the libcrypto included on your system,
```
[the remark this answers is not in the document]
TingDaoK: found a setting for my formatter to disable the extra space in between the sentence. 👍

test/test_s3.py
```diff
@@ -594,6 +625,74 @@ def on_done_remove_file(**kwargs):
...
+                # also wait for every thread to be joined, incase of some thread is in the middle of cleanup.
+                shutdown_event = CRT_S3_CLIENT.shutdown_event
+                CRT_S3_CLIENT = None
+                gc.collect()
+                self.assertTrue(shutdown_event.wait(self.timeout))
+                self.assertTrue(join_all_native_threads(timeout_sec=0.1))
+        except Exception as e:
+            print(f"before_fork error: {e}")
+            # fail hard
+            exit(-1)
```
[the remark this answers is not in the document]
TingDaoK: https://github.com/awslabs/aws-crt-python/actions/runs/13904462279/job/38904049353
looks like it flushes the output and makes the exception raised for unittest
but the process still hangs...

test/test_s3.py
```diff
@@ -594,6 +625,74 @@ def on_done_remove_file(**kwargs):
...
+        try:
+            if CRT_S3_CLIENT is not None:
+                # The client is not safe to use after fork, so we need to release it.
+                # make sure the client is shutdown properly before fork
+                # also wait for every thread to be joined, incase of some thread is in the middle of cleanup.
+                shutdown_event = CRT_S3_CLIENT.shutdown_event
+                CRT_S3_CLIENT = None
+                gc.collect()
+                self.assertTrue(shutdown_event.wait(self.timeout))
+                self.assertTrue(join_all_native_threads(timeout_sec=0.1))
```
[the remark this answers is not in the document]
TingDaoK: good find!

.github/workflows/ci.yml
```diff
@@ -65,23 +64,43 @@ jobs:
...
-      - name: Install qemu/docker
-        run: docker run --privileged --rm tonistiigi/binfmt --install aarch64
-
       - name: Build ${{ env.PACKAGE_NAME }}
         run: |
           aws s3 cp s3://aws-crt-test-stuff/ci/${{ env.BUILDER_VERSION }}/linux-container-ci.sh ./linux-container-ci.sh && chmod a+x ./linux-container-ci.sh
           ./linux-container-ci.sh ${{ env.BUILDER_VERSION }} aws-crt-manylinux2014-${{ matrix.image }} build -p ${{ env.PACKAGE_NAME }} --python /opt/python/${{ matrix.python }}/bin/python
 
+  manylinux2014-arm64:
+    runs-on: ubuntu-24.04-arm
```
graebm, line 73: 🤩

README.md
```diff
@@ -34,6 +34,35 @@ python3 -m pip install .
 
 To use from your Python application, declare `awscrt` as a dependency.
```
graebm, line 35: trivial: link the "advanced" options from up here in the "simple" build instructions...

Just replace this line about dependencies. Users that need to be told this, would also need us to show them how to do it, this isn't a library for users who need that much hand holding

```suggestion
See [Advanced Build Options](#advanced-build-options) for more info about building from source.
```

graebm (approved):
looking good. Just some questions about the arm64 runners

.github/workflows/ci.yml
TingDaoK: by default, builder doesn't run tests on arm https://github.com/awslabs/aws-crt-builder/blob/main/builder/core/data.py#L318-L329

But, we run our tests via the build action in `AWSCrtPython`.

So, it still runs the action, but mqtt stuff was skipped.

PR #629, opened 2025-03-13
Add pyproject.toml, and stop calling setup.py directly

graebm:
**Issue:**

For a few years now, we've been seen the occasional deprecation warning about setup.py. Including:

    ********************************************************************************
    Please avoid running ``setup.py`` directly.
    Instead, use pypa/build, pypa/installer or other
    standards-based tools.

    See https://blog.ganssle.io/articles/2021/10/setup-py-deprecated.html for details.
    ********************************************************************************

and

    ********************************************************************************
    Ensure that any custom bdist_wheel implementation is a subclass of
    setuptools.command.bdist_wheel.bdist_wheel.

    By 2025-Oct-15, you need to update your project and remove deprecated calls
    or your builds will no longer be supported.

    See https://github.com/pypa/wheel/pull/631 for details.
    ********************************************************************************

2025-Oct is on its way, so let's finally do something about it...

**Description of changes:**

Follow advice from ["Is setup.py deprecated?"](https://packaging.python.org/en/latest/discussions/setup-py-deprecated/) page at packaging.python.org

tldr; having a setup.py file **is not** deprecated, but directly invoking `python setup.py` **is** deprecated.

- Add [pyproject.toml](https://packaging.python.org/en/latest/specifications/pyproject-toml/) file (the modern way to configure your package)
    - Move data from `setup.py` -> `pyproject.toml` wherever possible (can't move dynamic stuff, like build steps for C extension)
    - Move dev dependencies from `requirements-dev.txt` -> `pyproject.toml [project.optional-dependencies]` 
        - Now you call `pip install ".[dev]"`, instead of `pip install -r requirements-dev.txt`. The dev dependencies will be installed by the same command that builds and installs our project.
        - This change isn't 100% necessary, but seemed like the modern way to do things. It's one less step to set things up on your machine. However, you can't install dependencies without ALSO building/installing the package (we can roll this back if we end up hating it)
- Use `python3 -m build` instead of `python3 setup.py sdist bdist_wheel` (the modern way to build a wheel for distribution)
    - Sadly, [build](https://pypi.org/project/build/) is a non-core package, so might need to be pip installed before using. Fortunately it's been [preinstalled on manylinux images since 2021](https://github.com/pypa/manylinux/blob/6b97f150fc9ede1cb4f25de84a939538a1019a58/docker/build_scripts/requirements3.8.txt#L7), so our Linux release process will be OK.
    - I updated our MacOS and Windows machines to have `build` installed for all python versions, so our release process should be OK.
- STOP calling `pip install --upgrade setuptools` before `pip install .`
    - Modern pip installs the "build dependencies" declared in `pyproject.toml` (setuptools & wheel) in a special isolated environment before it builds our project. We declared `setuptools>=75.3.1`, so we're guaranteed it will get installed, and be a recent version.


.github/workflows/ci.yml
```diff
@@ -337,8 +337,8 @@ jobs:
           submodules: true
       - name: Run tests
         run: |
-          python3 -m pip install --upgrade --requirement requirements-dev.txt
-          python3 -m pip install . --verbose
+          python3 -m pip install --upgrade pip
```
DmitriyMusatkin, line 340: semi related: do we still think upgrading pip is a good idea. i remember we had some issues in the past year with python blocking global pip upgrade in some cases and instead pointing to upgrading pip within venv

.github/workflows/lint.yml
```diff
@@ -7,28 +7,16 @@ on:
...
 
-    - name: Build and Test
+    - name: Format and Check Diff
       run: |
-        python3 -m pip install --upgrade setuptools
-        python3 -m pip install --upgrade wheel
         python3 -m pip install --upgrade autopep8
-        python3 -m autopep8 --exit-code --diff --recursive awscrt test .builder setup.py
+        python3 scripts/format-all.py
+        git diff
```
DmitriyMusatkin, line 21: do we want both?

DmitriyMusatkin (approved):

.github/workflows/ci.yml
graebm: I'll try taking these out and see what happens ... and it failed.

tried moving the job to ubuntu-24.04, which would have newer pip ... and that failed too due to formatting errors raised by sphinx.

The docs are sadly very fragile. Don't have time to diagnose. Just making it work for now :(

.github/workflows/lint.yml
graebm: I switched this to just use the formatting script we already have. But that script just changes the code in-place, it doesn't have a "raise error if different" mode

`git diff --quiet` has exit code 1 if anything has changed, but it doesn't print anything

so I'm doing `git diff` before that, to show what's different (but it always exits with 0) before that

graebm: I added comments to the file...

PR #635, opened 2025-03-31
H2 write stream

TingDaoK:
*Issue #, if available:*

- binding out the [aws_http2_stream_write_data](https://github.com/awslabs/aws-c-http/blob/main/include/aws/http/request_response.h#L930)
- Modify the interface to have http2 specific connection and stream leave the opportunity for us to adding more http/2 specific features.

*Description of changes:*
- In `aws_py_translate_py_error`, clearly don't `set_sys_last_vars`. Since the `set_sys_last_vars` will leave the exception (traceback) object live in the sys level, and somehow it may refer to the CRT native objects, and causing the test to leak. (which in theory, it's probably fine?)
- Adding new API to allow user to use `aws_http2_stream_write_data` to avoid every ticket of the loop to check the input stream for more data. It will wait until the next available input stream to read from it, which reduces the CPU usage.

**Key Design options**
1. The connection/stream interface is a low level details in terms of the http client. *If we implement the http client in the future, the connection/stream won't be the interface we want public usage*
2. Given the low level interface here, I believe we should have the separate interface for HTTP/1.1 and HTTP/2, we can build the unified interface on top of this in the future, but we cannot get the other way to leave the door open.
3. We add those functionalities based on the use case from python SDK, only add as needed.
4. Some good features like "allow the alpn to return different protocol" and "helper to config the http/2 for the http/2 connection" as not currently needed.


awscrt/http.py
```diff
@@ -112,6 +112,29 @@ def new(cls,
...
+            port,
+            bootstrap,
+            socket_options,
+            tls_connection_options,
+            proxy_options)
+
+    @classmethod
+    def _generic_new(
+            _cls,
+            expected_cls,
```
graebm, line 127: a `classmethod` is just a `staticmethod` that automatically passes `cls`
if you don't want automatic `cls`, do `staticmethod` instead
```suggestion
    @staticmethod
    def _generic_new(
            expected_cls,
```

awscrt/http.py
```diff
@@ -127,14 +150,18 @@ def new(cls,
...
+            connection = expected_cls()
             connection._host_name = host_name
             connection._port = port
 
             def on_connection_setup(binding, error_code, http_version):
                 if error_code == 0:
                     connection._binding = binding
                     connection._version = HttpVersion(http_version)
+                    if expected_version and expected_version != connection._version:
+                        # unexpected protocol version
```
graebm, line 162: Currently, `expected_version` is always None

graebm: If Http2ClientConnection.new() is always supposed to create an HTTP/2 connection, shouldn't it do things to try and guarantee that's the kind of connection which gets made? Like set the ALPN string, or prior_knowledge?

awscrt/http.py
```diff
@@ -127,14 +150,18 @@ def new(cls,
             if not bootstrap:
                 bootstrap = ClientBootstrap.get_or_create_static_default()
 
-            connection = cls()
+            connection = expected_cls()
```
graebm, line 153: we could always delay instantiating `connection` with the proper class type until we know the actual http_version, later in on_connection_setup

Like, we could go back to just having 1 HttpClientConnection.new() function, and you may end up with an Http2ClientConnection or HttpClientConnection instance, depending

awscrt/http.py
```diff
@@ -307,6 +370,27 @@ def _on_complete(self, error_code):
...
+class Http2ClientStream(HttpClientStream):
+    def __init__(self, connection, request, on_response=None, on_body=None, manual_write=False):
+        super()._generic_init(connection, request, on_response, on_body, manual_write)
+
+    def write_data(self, data_stream, end_stream=False):
+        future = Future()
+        if data_stream is not None:
+            body_stream = InputStream.wrap(data_stream)
+        else:
+            body_stream = None
```
graebm, line 382: utterly trivial: simpler
```suggestion
        data_stream = InputStream.wrap(data_stream, allow_none=True)
```

awscrt/http.py
```diff
@@ -126,24 +146,34 @@ def new(cls,
...
+                connection._host_name = host_name
+                connection._port = port
 
-            # on_shutdown MUST NOT reference the connection itself, just the shutdown_future within it.
-            # Otherwise we create a circular reference that prevents the connection from getting GC'd.
-            shutdown_future = connection.shutdown_future
+                connection._binding = binding
+                connection._version = HttpVersion(http_version)
+                nonlocal shutdown_future
+                shutdown_future = connection.shutdown_future
```
graebm, line 170: debatable: maybe move to the [_Core pattern](https://github.com/awslabs/aws-crt-python/blob/af06aa55085c437351b32a5870f48fd6a1320f97/guides/dev/writing-code.md#option-2---private-core-class)? this is getting pretty fancy

awscrt/http.py
```diff
@@ -219,6 +249,33 @@ def request(self, request, on_response=None, on_body=None):
...
 
 
+class Http2ClientConnection(HttpClientConnection):
+    """
+    HTTP/2 client connection.
+
+    This class extends HttpClientConnection with HTTP/2 specific functionality.
+    """
+    @classmethod
+    def new(cls,
```
graebm, line 259: debatable: not sure it's worth giving this an explicit constructor? It's not doing anything to actually *cause* the user to get an HTTP/2 connection, like setting alpn=h2 or prior_knowledge_http2. It's just forcing an error if the user didn't end up getting an HTTP/2 connection.

For better or worse, our API in C takes all possible options, because it's possible to establish a connection and not know what you're going to get.

awscrt/http.py
```diff
@@ -112,6 +112,26 @@ def new(cls,
...
+        return HttpClientConnection._generic_new(
+            host_name,
+            port,
+            bootstrap,
+            socket_options,
+            tls_connection_options,
+            proxy_options)
+
+    @staticmethod
+    def _generic_new(
```
DmitriyMusatkin, line 124: nit: ignore if this has been discussed already. generic name seems a bit weird. maybe base or common or something?

DmitriyMusatkin (approved):

source/http_stream.c
```diff
@@ -235,7 +236,8 @@ PyObject *aws_py_http_client_stream_new(PyObject *self, PyObject *args) {
     PyObject *py_stream = NULL;
     PyObject *py_connection = NULL;
     PyObject *py_request = NULL;
-    if (!PyArg_ParseTuple(args, "OOO", &py_stream, &py_connection, &py_request)) {
+    int h2_manual_write = 0;
```
waahm7, line 239:
```suggestion
    int http2_manual_write = 0;
```

PR #641, opened 2025-04-04
Override macOS deployment target

xiazhvera:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -11,7 +11,13 @@ import shutil
 import subprocess
 import sys
 import sysconfig
+from packaging import version
 from wheel.bdist_wheel import bdist_wheel
+
+# The MACOSX_DEPLOYMENT_TARGET for the library. If not set, it will use the
+# default value from python sysconfig.get_config_var('MACOSX_DEPLOYMENT_TARGET').
+MACOS_DEPLOYMENT_TARGET_OVERRIDE = "10.15"
+
 if sys.platform == 'win32':
     # distutils is deprecated in Python 3.10 and removed in 3.12. However, it still works because Python defines a compatibility interface as long as setuptools is installed.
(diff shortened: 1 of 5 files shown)


setup.py
```diff
@@ -12,6 +12,11 @@
 import sys
 import sysconfig
 from wheel.bdist_wheel import bdist_wheel
+
+# Override the default MACOSX_DEPLOYMENT_TARGET for the build. If not set, it will use the
+# default value from python sysconfig.get_config_var('MACOSX_DEPLOYMENT_TARGET').
+MACOS_DEPLOYMENT_TARGET_OVERRIDE = 10.15
+
```
waahm7, line 19: I think we need to set the environment variable so that pypi can pick it up. Setting it only for C code won't work.

setup.py
```diff
@@ -11,7 +11,13 @@
 import subprocess
 import sys
 import sysconfig
+from packaging import version
 from wheel.bdist_wheel import bdist_wheel
+
+# The MACOSX_DEPLOYMENT_TARGET for the library. If not set, it will use the
+# default value from python sysconfig.get_config_var('MACOSX_DEPLOYMENT_TARGET').
+MACOS_DEPLOYMENT_TARGET_OVERRIDE = "10.15"
```
graebm, line 19: utterly trivial/style: move this variable declaration down, so it happen after all the `import` statements

currently, it's right before a weird conditional `import`

maybe move it down near the AWS_LIBS declaration

graebm, line 19: It's not technically an "override", it's a "minimum". We'll end up using whichever value is higher
```
# Minimum MACOSX_DEPLOYMENT_TARGET needed to build AWS_LIBS
MACOS_DEPLOYMENT_TARGET_MIN = "10.15"
```

graebm (approved):

PR #642, opened 2025-04-07
Helper functions for non-optional integers should fail if the value is None

bretambrose:

### DIFF
--- a/source/module.c
+++ b/source/module.c
@@ -119,7 +119,7 @@ PyObject *PyUnicode_FromAwsString(const struct aws_string *aws_str) {
 uint32_t PyObject_GetAttrAsUint32(PyObject *o, const char *class_name, const char *attr_name) {
     uint32_t result = UINT32_MAX;
 
-    PyObject *attr = PyObject_GetAttrString(o, attr_name);
+    PyObject *attr = PyObject_GetAttrString(o, attr_name); /* new reference */
     if (!attr) {
         PyErr_Format(PyExc_AttributeError, "'%s.%s' attribute not found", class_name, attr_name);
         return result;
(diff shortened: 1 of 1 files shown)


sfod:
Shouldn't [PyObject_IsTrue](https://github.com/awslabs/aws-crt-python/blob/v0.25.7/source/module.c#L167-L173) and [PyObject_GetAttrAsIntEnum](https://github.com/awslabs/aws-crt-python/blob/v0.25.7/source/module.c#L194) also handle `Py_None` explicitly?

bretambrose:
> Shouldn't [PyObject_IsTrue](https://github.com/awslabs/aws-crt-python/blob/v0.25.7/source/module.c#L167-L173) and [PyObject_GetAttrAsIntEnum](https://github.com/awslabs/aws-crt-python/blob/v0.25.7/source/module.c#L194) also handle `Py_None` explicitly?

I updated PyObject_GetAttrAsBool and PyObject_GetAttrAsIntEnum

sfod (approved):

source/module.c
```diff
@@ -125,6 +125,11 @@ uint32_t PyObject_GetAttrAsUint32(PyObject *o, const char *class_name, const cha
         return result;
     }
 
+    if (attr == Py_None) {
+        PyErr_Format(PyExc_AttributeError, "'%s.%s' required integral attribute is None", class_name, attr_name);
+        return result;
```
graebm, line 130: attr is a [new reference](https://docs.python.org/3/c-api/object.html#c.PyObject_GetAttrString)

need to decref it. maybe:

```suggestion
            goto done;
```
and then later
```
done:
    Py_DECREF(attr);
    return result;
}
```

that's how the [old versions](https://github.com/awslabs/aws-crt-python/blob/b27e0a70973c2f62feec123886dc30ac027b296c/source/module.c#L145) of these functions used to do it

graebm: If you really want to be a champ. Edit 
```
PyObject *attr = PyObject_GetAttrString(o, attr_name);
```
in all these functions to be like:
```
PyObject *attr = PyObject_GetAttrString(o, attr_name); /* new reference */
```

I mean, that's what we advise in our [coding guidelines](https://github.com/awslabs/aws-crt-python/blob/main/guides/dev/writing-code.md#reference-counting-in-c) (which were written years after most of this C code)

bretambrose: But it's Py_None in the case you're worried about.

I could add a comment saying why there's no need to decref it.

graebm: But PyObject_GetAttrString() still increments the refcount on it. Seems weird to leak a reference on purpose because you think it probably doesn't matter. Vs just doing the normal refcount thing

PR #647, opened 2025-04-17
Request response streams

bretambrose:
Streaming operation support and tests

### DIFF
--- a/awscrt/mqtt_request_response.py
+++ b/awscrt/mqtt_request_response.py
@@ -13,6 +13,7 @@ from concurrent.futures import Future
 import _awscrt
 import collections.abc
 
+
 class SubscriptionStatusEventType(IntEnum):
     """
     The type of change to the state of a streaming operation subscription
@@ -70,6 +71,7 @@ SubscriptionStatusListener = Callable[[SubscriptionStatusEvent], None]
 Signature for a handler that listens to incoming publish events.
 """
 IncomingPublishListener = Callable[[IncomingPublishEvent], None]
+
 
 @dataclass
 class StreamingOperationOptions:
@@ -273,7 +275,8 @@ class Client(NativeResource):
                 event = IncomingPublishEvent(topic, payload)
                 options.incoming_publish_listener(event)
 
-        stream_binding = _awscrt.mqtt_request_response_client_create_stream(self._binding, options.subscription_topic_filter, on_subscription_status_event, on_incoming_publish_event)
+        stream_binding = _awscrt.mqtt_request_response_client_create_stream(
+            self._binding, options.subscription_topic_filter, on_subscription_status_event, on_incoming_publish_event)
 
         return StreamingOperation(stream_binding)
 
@@ -282,6 +285,7 @@ class StreamingOperation(NativeResource):
     """
     An operation that represents a stream of events broadcast to an MQTT topic
     """
+
     def __init__(self, binding):
         super().__init__()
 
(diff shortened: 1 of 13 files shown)


xiazhvera (approved):
LGTM. Just need verify 311 test works properly.

source/mqtt_request_response.c
```diff
@@ -481,11 +483,198 @@ PyObject *aws_py_mqtt_request_response_client_make_request(PyObject *self, PyObj
...
+    void *user_data) {
+
+    struct mqtt_streaming_operation_binding *stream_binding = user_data;
+
+    PyGILState_STATE state;
+    if (aws_py_gilstate_ensure(&state)) {
+        return;
+    }
+
+    PyObject *result = PyObject_CallFunction(
```
xiazhvera, line 543: This seems to be a behavior difference between python and other SDK. For CPP SDK, if `aws_mqtt_rr_client_operation_release` is called, no callback would be invoked anymore. 
https://github.com/awslabs/aws-crt-cpp/blob/d792acb017f5596eb4356adff436aa18adaa727f/source/iot/MqttRequestResponseClient.cpp#L182

However, for python binding, there is no validation to stream_binding->native. Therefore, as long as the termination callback is not called, the callbacks will still be invoked regardless the python operation object is already destroyed.

xiazhvera (changes requested):

source/mqtt_request_response.c
bretambrose: That's intentional; it's safe/fine for it to do so.

awscrt/mqtt_request_response.py
```diff
@@ -64,6 +150,9 @@ class RequestResponseOperationOptions:
     correlation_token: 'Optional[str]' = None
 
     def validate(self):
+        """
+        Stringently type-checks an instance's field values.
+        """
         assert isinstance(self.subscription_topic_filters, collections.abc.Sequence)
```
sbSteveK, line 156: out of scope of this specific review but could we use `from collections.abc import Sequence` in imports and use `Sequence` directly? I'm also getting "Sequence is not defined" in the areas of this file where `Sequence` is being used directly despite `import collections.abc` being up top.

source/mqtt_request_response.c
xiazhvera: My concern is more about the behavioral difference between SDKs rather than safety. In the CPP SDK, the callback is no longer invoked once the client is shut down, while in python, callbacks may still be triggered even after the client is destroyed.

I’m fine with either behavior, but it would be better if we keep them consistent across SDKs.

source/mqtt_request_response.c
```diff
@@ -481,11 +483,198 @@ PyObject *aws_py_mqtt_request_response_client_make_request(PyObject *self, PyObj
...
+    if (!stream_binding) {
+        return NULL;
+    }
+
+    if (aws_mqtt_rr_client_operation_activate(stream_binding->native)) {
+        PyErr_SetAwsLastError();
+        return NULL;
+    }
+
+    return Py_None;
```
sbSteveK, line 661: This should use `return Py_RETURN_NONE` instead of `return Py_None`. `Py_None` is ref-counted and its ref-count will be decremented after ` _awscrt.mqtt_streaming_operation_open(self._binding)` is done with the result. `Py_RETURN_NONE` increments the ref of Py_None before returning it. Not incrementing the ref can result in either a deallocating None error or the memory being funky and reused in weird, potentially bad ways.

xiazhvera (approved):

PR #654, opened 2025-05-13
Fix version str, strip the leading v

TingDaoK:
*Issue #, if available:*
- Forgot to strip the leading `v` when update the version string.

*Description of changes:*


continuous-delivery/update-version.py
```diff
@@ -10,6 +10,9 @@
 # convert v0.2.12-2-g50254a9 to 0.2.12
 # test-version-exists will ensure to not include non-tagged commits
 version = tag.split('-', 1)[0]
+# strip the leading v
+version = str(tag[1:].strip(), 'utf8')
```
waahm7, line 14: Can you assert that it's a v before stripping? If someone forgets v in the GitHub tag, this will end up with a botched version string.

TingDaoK: added the check to `test-version-exists` script

DmitriyMusatkin (approved):

PR #658, opened 2025-06-11
Asyncio

TingDaoK:
*Issue #, if available:*

*Description of changes:*


awscrt/http_asyncio.py
JordonPhillips, line 1: There's no "official" way to separate out async versions of tools, but consensus seems to be going in the direction of having an `aio` module with the async versions contained there in similar structure to the sync components.

awscrt/http_asyncio.py
```diff
@@ -0,0 +1,337 @@
...
+        """Close the connection asynchronously.
+
+        Shutdown is asynchronous. This call has no effect if the connection is already
+        closing.
+
+        Returns:
+            None: When shutdown is complete.
+        """
+        close_future = super().close()
+        await asyncio.wrap_future(close_future)
```
JordonPhillips, line 85: Generally you shouldn't change the type of an operation in a subclass. I would recommend either exposing a new method entirely (e.g. `close_async` or something) or you could just not subclass and instead wrap the sync client connection.

awscrt/http_asyncio.py
```diff
@@ -0,0 +1,337 @@
...
+                connection options. If None is provided, then the connection will
+                be attempted over plain-text.
+
+            proxy_options (Optional[HttpProxyOptions]): Optional proxy options.
+                If None is provided then a proxy is not used.
+
+        Returns:
+            HttpClientConnectionAsync: A new HTTP client connection.
+        """
+        future = HttpClientConnection._generic_new(
```
JordonPhillips, line 65: You shouldn't reference the private method of another class. The following will achieve the same thing while referencing this class's version of the method:

```suggestion
        future = cls._generic_new(
```

awscrt/aio/aiohttp.py
```diff
@@ -0,0 +1,436 @@
...
+            else:
+                future.set_result(None)
+
+        _awscrt.http2_client_stream_write_data(self, body_stream, end_stream, on_write_complete)
+        await asyncio.wrap_future(future)
+
+    async def _set_request_body_generator(self, body_iterator: AsyncIterator[bytes]):
+        try:
+            async for chunk in body_iterator:
+                await self._write_data(io.BytesIO(chunk), False)
```
JordonPhillips, line 432: Having to wrap this in a BytesIO is really unfortunate as that creates a copy of the bytes passed in, which is then presumably copied again when it makes it's way to the CRT. Can we not write the bytes directly? Or at least provide an `InputStream` implementation that takes those bytes directly and transfers them without creating an intermediate copy? Having to construct a new `InputStream` every time is also not amazing since that's a bunch of allocations, but at least that should be comparatively tiny.

awscrt/aio/aiohttp.py
```diff
@@ -0,0 +1,436 @@
...
+        except Exception:
+            raise
```
JordonPhillips, line 434: This does nothing, so should be removed

awscrt/aio/aiohttp.py
```diff
@@ -0,0 +1,436 @@
...
+    def __init__(self,
+                 connection: AIOHttpClientConnection,
+                 request: HttpRequest,
+                 request_body_generator: AsyncIterator[bytes] = None,
+                 loop: Optional[asyncio.AbstractEventLoop] = None) -> None:
+        super().__init__(connection, request, request_body_generator=request_body_generator, loop=loop)
+
+    async def _write_data(self, body, end_stream):
+        future: Future = Future()
+        body_stream: InputStream = InputStream.wrap(body, allow_none=True)
```
JordonPhillips, line 415: This isn't a blocker, but you don't need to annotate types on assignment unless the type you want to declare is wider than the type being assigned. For example:

```python
optional_message: str | None = None
if self.message_available():
    optional_message = self.read_message()
```

Another example of when you'd do this if you want to widen parametric types on a generic.

awscrt/aio/aiohttp.py
```diff
@@ -0,0 +1,436 @@
...
+    async def wait_for_completion(self) -> int:
+        """Wait asynchronously for the stream to complete.
+
+        Returns:
+            int: The response status code.
+        """
+        return await self._completion_future
+
+    async def _set_request_body_generator(self, body_iterator: AsyncIterator[bytes]):
+        ...
```
JordonPhillips, line 366: Is HTTP1.1 not able to share the implementation?

awscrt/aio/aiohttp.py
TingDaoK: It should already avoid the extra copy, it will be wrapped by our input stream interface use `readinto`.
https://github.com/awslabs/aws-crt-python/blob/main/awscrt/io.py#L689-L727

awscrt/aio/aiohttp.py
```diff
@@ -0,0 +1,434 @@
...
+    """
+
+    @classmethod
+    async def new(cls,
+                  host_name: str,
+                  port: int,
+                  bootstrap: Optional[ClientBootstrap] = None,
+                  socket_options: Optional[SocketOptions] = None,
+                  tls_connection_options: Optional[TlsConnectionOptions] = None,
+                  proxy_options: Optional['HttpProxyOptions'] = None) -> "AIOHttpClientConnectionUnified":
```
DmitriyMusatkin, line 41: why does HttpProxyOptions need to be forward declared?

awscrt/aio/aiohttp.py
```diff
@@ -0,0 +1,434 @@
...
+            if future.cancelled():
+                # the future was cancelled, so we don't need to set the result or exception
+                return
+            if error_code:
+                future.set_exception(awscrt.exceptions.from_code(error_code))
+            else:
+                future.set_result(None)
+
+        _awscrt.http2_client_stream_write_data(self, body_stream, end_stream, on_write_complete)
+        await asyncio.wrap_future(future)
```
DmitriyMusatkin, line 427: why do we want to use concurrent Future here and then wrap it into asyncio one? why not use async io one from start?

DmitriyMusatkin (approved):

awscrt/aio/aiohttp.py
TingDaoK: Since the concurrent future is thread safe, but the asyncio one is not.

If we just get asyncio future, we will need to invoke `self._loop.call_soon_threadsafe()` to set the future. Either way, it's a bit awkward.

But, yeah, I did use the asyncio future directly for some other places, I guess it would be nice to keep it consistent.

dacevedo12:
just out of curiosity, why aiohttp instead of httpx?

awscrt/http.py
```diff
@@ -337,17 +343,74 @@ def request(self,
                 on_response: Optional[Callable[..., None]] = None,
                 on_body: Optional[Callable[..., None]] = None,
                 manual_write: bool = False) -> 'Http2ClientStream':
+        """Create `Http2ClientStream` to carry out the request/response exchange.
+
+        NOTE: The HTTP stream sends no data until `Http2ClientStream.activate()`
```
nateprewitt, line 348: What is the envisioned lifecycle for a stream? Are we activating and closing every transaction or do we send many requests?

awscrt/aio/aiohttp.py
```diff
@@ -0,0 +1,423 @@
...
+                  bootstrap: Optional[ClientBootstrap] = None,
+                  socket_options: Optional[SocketOptions] = None,
+                  tls_connection_options: Optional[TlsConnectionOptions] = None,
+                  proxy_options: Optional[HttpProxyOptions] = None,
+                  initial_settings: Optional[List[Http2Setting]] = None,
+                  on_remote_settings_changed: Optional[Callable[[List[Http2Setting]],
+                                                                None]] = None) -> "AIOHttp2ClientConnection":
+        """
+        Asynchronously establish an HTTP/2 client connection.
+        Notes: to set up the connection, the server must support HTTP/2 and TlsConnectionOptions
```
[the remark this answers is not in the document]
TingDaoK: this will start the http connection right away.
It's a connection interface, it won't handle any of the reconnect logic or reuse the connection.

I assume the connection manager, is what you are expecting to about `how to connect`. https://github.com/awslabs/aws-c-http/blob/main/include/aws/http/connection_manager.h

awscrt/aio/aiohttp.py
```diff
@@ -0,0 +1,423 @@
...
+        """
+        return AIOHttpClientStreamUnified(self, request, request_body_generator, loop)
+
+
+class AIOHttpClientConnection(AIOHttpClientConnectionUnified):
+    """
+    An async HTTP/1.1 only client connection.
+
+    Use `AIOHttpClientConnection.new()` to establish a new connection.
+    """
```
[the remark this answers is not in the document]
TingDaoK: `AIOHttpClientConnection` is designed to be HTTP/1.1 only. the protocol doesn't support multiplex, but we do some sort support it. The next request/response will just be waiting in the list to be sent after the previous request/response finishes. 

>  The request() method is also not async

There is no async process in the `request()` method, it will start the request under the hood, and the stream interface will be handle the response async.

PR #694, opened 2025-10-30
Change the loglevel of logger

sbSteveK:
This exposes `aws_logger_set_log_level` for use in aws-crt-python via `io.set_log_level()`. This will allow customers to change the loglevel of logs after they have created a logger.

### DIFF
--- a/awscrt/io.py
+++ b/awscrt/io.py
@@ -41,7 +41,7 @@ def init_logging(log_level, file_name):
 
 def set_log_level(log_level):
     """Change the log level of `awscrt`. init_logging() must have been called
-    before using this function.
+    before using this function or else an exception will be raised.
 
     Args:
         log_level (LogLevel): Display messages of this importance and higher.


awscrt/io.py
```diff
@@ -40,6 +39,18 @@ def init_logging(log_level, file_name):
     _awscrt.init_logging(log_level, file_name)
 
 
+def set_log_level(log_level):
+    """Change the log level of `awscrt`. init_logging() must have been called
+    before using this function.
```
TingDaoK: Or else, exception will be raised

source/module.c
```diff
@@ -86,6 +86,27 @@ PyObject *aws_py_init_logging(PyObject *self, PyObject *args) {
...
+
+    if (!s_logger_init) {
+        aws_raise_error(AWS_ERROR_INVALID_STATE);
+        return PyErr_AwsLastError();
+    }
+
+    int log_level = 0;
+    if (!PyArg_ParseTuple(args, "b", &log_level)) {
+        PyErr_SetNone(PyExc_ValueError);
+        return NULL;
```
TingDaoK: raise error?

TingDaoK: Oh, the python raises error by itself.

PR #700, opened 2025-11-21
Checks if futures are done in AIOHttpClientStreamUnified to avoid race condition

nicolastomeo:
*Issue #, if available:*
In Python libraries like https://github.com/awslabs/aws-sdk-python/tree/develop/clients/aws-sdk-transcribe-streaming and https://github.com/awslabs/amazon-transcribe-streaming-sdk where awscrt is used, a few people, me included notice an issue where futures that are already done (cancelled) are being set so an InvalidStateError is raised.
Examples: https://github.com/awslabs/amazon-transcribe-streaming-sdk/issues/61 and https://github.com/awslabs/aws-sdk-python/issues/13

*Description of changes:*
Just checking if future is not done before setting the result.

### DIFF
--- a/awscrt/aio/http.py
+++ b/awscrt/aio/http.py
@@ -295,7 +295,8 @@ class AIOHttpClientStreamUnified(HttpClientStreamBase):
         """Process body chunk on the correct event loop thread."""
         if self._chunk_futures:
             future = self._chunk_futures.popleft()
-            future.set_result(chunk)
+            if not future.done():
+                future.set_result(chunk)
         else:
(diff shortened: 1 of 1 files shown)


nicolastomeo:
Tagging @jonathan343 since I saw him actively working on https://github.com/awslabs/aws-sdk-python

Thanks in advance.

TingDaoK:
@nicolastomeo 
Hi, thank you for submitting the PR.
Do you have a reproduce case that we can use to locate the state better and have a better understanding of this issue?

PR #710, opened 2025-12-23
Add IoT Metrics Support 

xiazhvera:
*Issue #, if available:*

*Description of changes:*
- Bind `IoTDeviceSDKMetrics` and metrics related features , CRT is now appending AWS metrics by default.
- Add `enableMetrics` option to allow user disable metrics.

### DIFF
--- a/awscrt/_aws_iot_metrics.py
+++ b/awscrt/_aws_iot_metrics.py
@@ -3,6 +3,7 @@
 
 from dataclasses import dataclass
 
+
 @dataclass
 class AWSIoTMetrics:
     """


source/mqtt_client_connection.c
```diff
@@ -773,6 +808,13 @@ PyObject *aws_py_mqtt_client_connection_connect(PyObject *self, PyObject *args)
         }
     }
 
+    // If metrics is None, we do not set metrics at all.
+    if (metrics_py != Py_None) {
+        if (!s_set_metrics(py_connection->native, metrics_py)) {
+            goto done;
```
sfod: In crt-cpp, a failure on setting metrics is ignored. Here, it fails connection attempts. I think we should document the correct behavior and implement it everywhere.

sfod: There is still inconsistency on setting metrics failure between crt-cpp (just log failure and proceed with connection) and crt-python (fail connect attempt on failure). I think the cpp logic is the correct one because we don't want the metrics-related issues to break users.

awscrt/mqtt.py
```diff
@@ -330,6 +330,10 @@ class Connection(NativeResource):
 
         proxy_options (Optional[awscrt.http.HttpProxyOptions]):
             Optional proxy options for all connections.
+
+        enable_metrics (bool): If true, enable IoT SDK metrics in CONNECT packet username field, otherwise, disabled.
```
sbSteveK: Trivial: I think we should work on the wording here. The initial sentence makes it sound like the default is false. Instead of the "If true" maybe we should just describe what the bool is for. "Enable or disable IoT SDK metrics in CONNECT packet username field." Also wondering if we should add a bit of extra text regarding why but that's debatable. e.g. "Our team uses these metrics to help determine what features and improvements to prioritize." We can leave the "Default to True" but I would remove the follow-up that provides a reason to set it to false.

awscrt/mqtt5.py
```diff
@@ -17,6 +17,18 @@
 from inspect import signature
 
 
+@dataclass
+class SdkMetrics:
+    """
+    Configuration for IoT SDK metrics that are embedded in MQTT username field.
```
sbSteveK: Trivial/debatable: MQTT Connect Packet username field ?

awscrt/mqtt.py
```diff
@@ -15,7 +15,7 @@
 from awscrt.http import HttpProxyOptions, HttpRequest
 from awscrt.io import ClientBootstrap, ClientTlsContext, SocketOptions
 from dataclasses import dataclass
-from awscrt.mqtt5 import Client as Mqtt5Client
+from awscrt.mqtt5 import Client as Mqtt5Client, SdkMetrics
```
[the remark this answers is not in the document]
sfod: We could add a separate module, like we did with req-resp client. With meta data implemented, the metrics functionality should be big enough to be in its own module.

awscrt/mqtt5.py
```diff
@@ -1715,6 +1722,7 @@ def __init__(
         self.keep_alive_secs: int = 1200 if keep_alive_secs is None else keep_alive_secs
         self.ack_timeout_secs: int = 0 if ack_timeout_secs is None else ack_timeout_secs
         self.clean_session: bool = True if clean_session is None else clean_session
+        self.enable_metrics: bool = True if enable_metrics is None else enable_metrics
```
sbSteveK, line 1725: Trivial: I think at this point it's impossible for enable_metrics to be None (Always defaults to true at creation) and can be set to enable_metrics.

sbSteveK (approved):
Seems solid. It looks like MQTT3 and MQTT5 will have to go through different binding paths when metrics is expanded. Would be ideal to use the same for both but unsure if that's possible without changes underneath?

xiazhvera:
> Seems solid. It looks like MQTT3 and MQTT5 will have to go through different binding paths when metrics is expanded. Would be ideal to use the same for both but unsure if that's possible without changes underneath?

Unfortunately, I don't think we can change the binding path for MQTT3 and MQTT5 without changes underneath...

PR #735, opened 2026-05-18
Metric update

rakshil14-2:
*Issue*

*Description of changes:*

- Made `AWSIoTMetrics` class as public. This class will hold all information regarding metrics.
- Add metrics feature encoding for each protocol.
- Added Metric Options so that user can set own metrics
- `enableMetrics` -> `disableMetrics` : change name convention to disable to focus on the "opt-out" pattern.


awscrt/aws_iot_metrics.py
```diff
@@ -0,0 +1,480 @@
...
+    metadata_entries: Optional[List[IoTMetricsMetadata]] = None
+
+
+# Metrics Version Constant
+IOT_SDK_METRICS_FEATURE_VERSION = 1
+
+# Feature ID Constants
+
+
+class MetricsFeatureId(str, Enum):
```
xiazhvera: I think we should keep the ids private. I didn't see why end user would need to use or know about the metrics value. Similar for the other encoding value classes.

awscrt/aws_iot_metrics.py
```diff
@@ -0,0 +1,480 @@
...
+            if entry.key == "IoTSDKMetricsVersion":
+                user_metrics_version = entry.value
+            elif entry.key == "IoTSDKFeature":
+                user_feature = entry.value
+            elif entry.key != "CRTVersion":
+                metadata[entry.key] = entry.value
+
+    # Merge features: if version matches, merge CRT + SDK; otherwise CRT only
+    if (user_metrics_version is not None and user_metrics_version.isdigit() and int(
+            user_metrics_version) == IOT_SDK_METRICS_FEATURE_VERSION and user_feature):
```
xiazhvera: It seems unnecessary to validate `user_features`, as the default `user_feature` will be "".

awscrt/aws_iot_metrics.py
```diff
@@ -0,0 +1,480 @@
...
+        metadata["IoTSDKFeature"] = _merge_feature_lists(crt_feature_list, "")
+
+    # Always set current metrics version
+    metadata["IoTSDKMetricsVersion"] = str(IOT_SDK_METRICS_FEATURE_VERSION)
+
+    final_metrics.metadata_entries = [IoTMetricsMetadata(key=k, value=v) for k, v in metadata.items()]
+    return final_metrics
+
+
+def create_metrics_mqtt5(client_options):
```
xiazhvera: Similarly, `create_metrics` should be a private APIs.

awscrt/aws_iot_metrics.py
```diff
@@ -0,0 +1,480 @@
...
+    for pair in crt_features.split(","):
+        if "/" in pair:
+            fid, val = pair.split("/", 1)
+            merged[fid] = val
+
+    for pair in user_features.split(","):
+        if "/" in pair:
+            fid, val = pair.split("/", 1)
+            merged[fid] = val
+    return ",".join(f"{k}/{v}" for k, v in sorted(merged.items()))
```
xiazhvera: I realize we also sorted the feature list in Swift, however, thinking about it, it seems unnecessary  to sort the feature list.

awscrt/io.py
```diff
@@ -606,12 +606,16 @@ class ClientTlsContext(NativeResource):
     Args:
         options (TlsContextOptions): Configuration options.
     """
-    __slots__ = ()
+    __slots__ = ('min_tls_ver', 'cipher_pref')
```
xiazhvera: As the members are for metrics tracking only, they should not be publicly accessible. Or it should at least be read-only. 
Swift made the same mistake.

awscrt/mqtt.py
```diff
@@ -412,8 +415,8 @@ def __init__(self,
         self.password = password
         self.socket_options = socket_options if socket_options else SocketOptions()
         self.proxy_options = proxy_options if proxy_options else websocket_proxy_options
-        if enable_metrics:
-            self._metrics = AWSIoTMetrics()
+        if not disable_metrics:
```
xiazhvera: Nit: Here is a double negative (`not disable_metrics`), consider flipping the condition for readability.

awscrt/mqtt5.py
```diff
@@ -1819,8 +1821,8 @@ def __init__(self, client_options: ClientOptions):
             socket_options = SocketOptions()
 
         # Handle metrics configuration
-        if client_options.enable_metrics:
-            self._metrics = AWSIoTMetrics()
+        if not client_options.disable_metrics:
```
xiazhvera: Nit: Similar, double negative (`not disable_metrics`) here.

source/mqtt5_client.c
```diff
@@ -1321,8 +1323,47 @@ PyObject *aws_py_mqtt5_client_new(PyObject *self, PyObject *args) {
     /* METRICS */
     struct aws_mqtt_iot_metrics metrics_tmp;
     AWS_ZERO_STRUCT(metrics_tmp);
-    if (PyObject_IsTrue(is_metrics_enabled_py)) {
-        metrics_tmp.library_name = metrics_library_name;
+
+    if (PyObject_IsTrue(is_metrics_enabled_py) && metrics_py != Py_None) {
```
xiazhvera, line 1368: Could we have a helper function to handle metrics object, and reuse it across mqtt5 and mqtt3?

xiazhvera (changes requested):

awscrt/aws_iot_metrics.py
```diff
@@ -0,0 +1,480 @@
...
+    SOCKET_IMPLEMENTATION = "G"
+    HTTP_PROXY_TYPE = "H"
+    CERTIFICATE_SOURCE = "I"
+    TLS_CIPHER_PREFERENCE = "J"
+    MINIMUM_TLS_VERSION = "K"
+
+# Feature Value Constants
+
+
+class _MetricsProtocolVersionValue(str, Enum):
```
xiazhvera: The naming conversion is different from `_retry_jitter_metrics_value` and other options below. It would be better to unify the naming.

awscrt/mqtt5.py
```diff
@@ -1819,10 +1821,12 @@ def __init__(self, client_options: ClientOptions):
             socket_options = SocketOptions()
 
         # Handle metrics configuration
-        if client_options.enable_metrics:
-            self._metrics = AWSIoTMetrics()
-        else:
+        enable_metrics = True
```
xiazhvera: Adding an extra variable `enable_metrics` seems unnecessary. Could we update the native function to use `disable_metrics` instead of enable_metrics?

source/mqtt5_client.c
```diff
@@ -831,6 +831,67 @@ static bool s_py_topic_aliasing_options_init(
...
+ * Metrics Parsing
+ ******************************************************************************/
+
+bool aws_py_metrics_parse(PyObject *metrics_py, struct aws_mqtt_iot_metrics *out_metrics) {
+    AWS_ZERO_STRUCT(*out_metrics);
+
+    PyObject *library_name_py = PyObject_GetAttrString(metrics_py, "library_name");
+    out_metrics->library_name = aws_byte_cursor_from_pyunicode(library_name_py);
+    Py_XDECREF(library_name_py);
+    if (!out_metrics->library_name.ptr) {
```
xiazhvera: Empty library name is a valid case. Similar for keys and values. 

Debatable, but I think we should not throw errors on metrics failure, because metrics is not necessary a user input.

source/mqtt5_client.c
```diff
@@ -831,6 +831,67 @@ static bool s_py_topic_aliasing_options_init(
     return success;
 }
 
+/*******************************************************************************
+ * Metrics Parsing
+ ******************************************************************************/
+
+bool aws_py_metrics_parse(PyObject *metrics_py, struct aws_mqtt_iot_metrics *out_metrics) {
+    AWS_ZERO_STRUCT(*out_metrics);
```
xiazhvera, line 839: The function header/comments should clearly state that the function expects an uninitialized struct and will zero it out.
If the caller has already allocated memory for `out_metrics` before calling this function, zeroing out the struct will cause a memory leak. It has to be explicitly call out that the function expect a uninitialized `out_metrics`.

xiazhvera (changes requested):

source/mqtt5_client.c
```diff
@@ -831,6 +831,67 @@ static bool s_py_topic_aliasing_options_init(
...
+    }
+
+    PyObject *metadata_entries_py = PyObject_GetAttrString(metrics_py, "metadata_entries");
+
+    if (metadata_entries_py && metadata_entries_py != Py_None && PyList_Check(metadata_entries_py)) {
+        Py_ssize_t count = PyList_Size(metadata_entries_py);
+        if (count > 0) {
+            struct aws_mqtt_metadata_entry *entries =
+                aws_mem_calloc(aws_py_get_allocator(), (size_t)count, sizeof(struct aws_mqtt_metadata_entry));
+            if (!entries) {
```
xiazhvera: We do not check `aws_mem_calloc()` result anymore. As OOM is unrecoverable and should abort.

xiazhvera (approved):

PR #757, opened 2026-07-22
Fix NASM failure against aws-lc

azkrishpy:
*Issue #, if available:*
nasm's Codeview debug-info backend (output/codeview.c) panics when it's asked to emit debug tables (-g flag, which the failing CI command uses) for an object with no code section — which is exactly what [aesni-xts-avx512.asm](https://github.com/aws/aws-lc/blob/main/generated-src/win-x86_64/crypto/fipsmodule/aesni-xts-avx512.asm) produces on win64 when MY_ASSEMBLER_IS_TOO_OLD_FOR_512AVX is defined.

This issue was observed since https://github.com/aws/aws-lc/pull/3319 which optimizes avx512 out when OPENSSL_SMALL is set (which we do by default).

*Description of changes:*
Bumped to the oldest version that actually fixes this issue. 

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -273,7 +273,7 @@ jobs:
           architecture: ${{ matrix.arch }}
       - uses: ilammy/setup-nasm@v1
         with:
-          version: '2.16.03'
+          version: '3.02rc7'
       - name: configure AWS credentials (containers)
         uses: aws-actions/configure-aws-credentials@v4
         with:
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -298,6 +300,8 @@ jobs:
           python-version: '3.8.10'
           architecture: ${{ matrix.arch }}
       - uses: ilammy/setup-nasm@v1
+        with:
+          version: '3.02rc8'
```
azkrishpy, line 304: should we use a tagged release or the oldest release with the fix? I don't have particular opinions on this.

azkrishpy:
the asm file will be fixed on aws-lc

PR #764, opened 2026-09-09
S3 Retry Option Bindings

sbSteveK:
Add `S3RetryConfig` dataclass to s3.py as an optional S3 client configuration allowing the assignment of the following members:
max_retries
backoff_scale_factor_ms
max_backoff_secs
jitter_mode
initial_bucket_capacity

Any options that are not set will use their S3 default values.

Moved `ExponentialBackoffJitterMode` from mqtt5.py into `io.py` as it is now shared across both mqtt and s3

### DIFF
--- a/awscrt/mqtt5.py
+++ b/awscrt/mqtt5.py
@@ -17,6 +17,7 @@ from collections.abc import Sequence
 from inspect import signature
 from awscrt.aws_iot_metrics import AWSIoTMetrics, _create_metrics_mqtt5
 
+
 class QoS(IntEnum):
     """MQTT message delivery quality of service.
 
@@ -795,6 +796,7 @@ class ClientOperationQueueBehaviorType(IntEnum):
     All operations that are not complete at the time of disconnection are failed, except operations that
     the MQTT5 spec requires to be retransmitted (un-acked QoS1+ publishes).
     """
+
 
 @dataclass
 class UserProperty:
--- a/awscrt/mqtt5.py
+++ b/awscrt/mqtt5.py
@@ -11,12 +11,11 @@ from concurrent.futures import Future
 from enum import IntEnum
 from awscrt import NativeResource, exceptions
 from awscrt.http import HttpProxyOptions, HttpRequest
-from awscrt.io import ClientBootstrap, SocketOptions, ClientTlsContext
+from awscrt.io import ClientBootstrap, SocketOptions, ClientTlsContext, ExponentialBackoffJitterMode
 from dataclasses import dataclass
 from collections.abc import Sequence
 from inspect import signature
 from awscrt.aws_iot_metrics import AWSIoTMetrics, _create_metrics_mqtt5
-
 
 class QoS(IntEnum):
     """MQTT message delivery quality of service.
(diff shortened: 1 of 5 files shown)


awscrt/mqtt5.py
```diff
@@ -798,35 +798,6 @@ class ClientOperationQueueBehaviorType(IntEnum):
     """
 
 
-class ExponentialBackoffJitterMode(IntEnum):
```
TingDaoK, line 801: hmmm, this is an API breaking change. you could probably do this in a non-breaking way, by point this to the io one, i think?

awscrt/s3.py
```diff
@@ -197,6 +197,43 @@ class S3FileIoOptions:
     """
 
 
+@dataclass
+class S3RetryConfig:
```
TingDaoK, line 201: why this belongs to s3? Should it be in io?

awscrt/mqtt5.py
sbSteveK: I don't think so, we still import it up top so anyone who imports mqtt5.py should still have access. Will check and confirm this is the case.

awscrt/s3.py
```diff
@@ -360,6 +401,16 @@ def on_shutdown():
...
             max_active_connections_override = 0
+
+        # Retry config: None -> 0 means "use S3 client defaults"
+        if retry_config is None:
+            retry_config = S3RetryConfig()
+        retry_max_retries = retry_config.max_retries if retry_config.max_retries is not None else 0
+        retry_backoff_scale_factor_ms = retry_config.backoff_scale_factor_ms if retry_config.backoff_scale_factor_ms is not None else 0
+        retry_max_backoff_secs = retry_config.max_backoff_secs if retry_config.max_backoff_secs is not None else 0
+        retry_jitter_mode = int(retry_config.jitter_mode) if retry_config.jitter_mode is not None else 0
+        retry_initial_bucket_capacity = retry_config.initial_bucket_capacity if retry_config.initial_bucket_capacity is not None else 0
```
TingDaoK: why not just default them to 0 instead of none? we are converting none to 0 here anyway.

awscrt/s3.py
sbSteveK: It's S3 specific. At least it's implementation as a struct is in aws-c-s3 within its config options.

awscrt/s3.py
sbSteveK: Remnant from MQTT way of handling optionals >,<. I'll default to 0 and just pass the value.

awscrt/s3.py
TingDaoK: but the retry config is not s3 specific, it was used for credentials provider as well

awscrt/mqtt5.py
sbSteveK: top level import keeps it safe and usable. Moving it to be shared across two areas that need it is the right path vs. having two of the same enums in two separate places with the same class name. They'd still be separate `s3.ExponentialBackoffJitterMode` `mqtt5.ExponentialBackoffJitterMode` but feels bad.

awscrt/s3.py
TingDaoK: ah... you created a retry config within s3 config instead of the `aws_exponential_backoff_retry_options`. okay...

TingDaoK (approved):

